PR #8, opened 2012-12-05
Add more efficient bpvec implementations for Snow and Dopar params

DarwinAwardWinner:
As discussed, I implemented proper splitting for both the Dopar and Snow backends. These implementations don't send the full object to every worker, and they also don't ever create the full split object.


sanmai-NL:
Can anyone explain why this PR is still open after such a long time?

mtmorgan:
Stale issue closed after remaining open for inexplicable length of time. The software has moved a lot from here; if there are new features / pull requests to address residual issues feel free to post.

PR #27, opened 2013-11-11
pass args via ... to mclapply

mllg:
Maybe not the cleanest solution possible, but it should be possible to pass additional arguments to mclapply.

### DIFF
--- a/R/MulticoreParam-class.R
+++ b/R/MulticoreParam-class.R
@@ -5,7 +5,8 @@
       recursive="logical",
       cleanup="logical",
       cleanupSignal="integer",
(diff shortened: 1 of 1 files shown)


sanmai-NL:
Can anyone explain why this PR is still open after such a long time?

mtmorgan:
Stale issue closed after remaining open for inexplicable length of time. The software has moved a lot from here; some components of this fix are not necessary any more, and the design (supporting all mclapply args via ...) is not consistent with BiocParallel vision. Ff there are new features / pull requests to address residual issues feel free to post.

PR #86, opened 2018-10-25
Supress slurm queue printing

grimbough:
Every time you run `BatchtoolsParam(cluster = "slurm")` it runs the system command `squeue` which for me seems to output to `stdout` rather than `stderr`, and results in thousands of lines printed in the R console.  This patch just redirects stdout to NULL.  It might be useful for other cluster types too.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: BiocParallel
 Type: Package
 Title: Bioconductor facilities for parallel evaluation
-Version: 1.17.0
+Version: 1.17.1
 Authors@R: c(
     person("Bioconductor Package Maintainer",
         email="maintainer@bioconductor.org", role="cre"),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: BiocParallel
 Type: Package
 Title: Bioconductor facilities for parallel evaluation
-Version: 1.15.15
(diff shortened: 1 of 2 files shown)


mtmorgan:
@nturaga can you review this and check in with me? In particular does the output still go _somewhere_, is it worth-while to make this an option of some sort, and do the other batchtools back-ends need similar consideration?

nturaga:
Hi @mtmorgan I'll take a look at this and get back to you.

DarwinAwardWinner:
It looks like the code in question is just running the `squeue` command to determine whether it is available and executable or not, which means only the exit code is relevant, and the output should indeed be discarded. The same is  probably true of all the other calls to `system2` in that function.

grimbough:
I concluded the same as @DarwinAwardWinner For reference the output looks like the below, but has no bearing on setting the parameters.
```
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
          23055241       gpu GMX_Benc    pchen  R       3:11      1 gpu2
```
I couldn't find anything definitive to say `squeue` always sends this to stdout, so stderr redirection may be sufficient in other instances, but stdout seems more natural.

I am happy to help test things that relate to SLURM if that's useful, but I don't have access to other HPC schedulers unfortunately.

nturaga:
Thanks @grimbough. This looks good.

PR #109, opened 2019-12-15
FIX: Partial argument name

HenrikBengtsson:
Fix avoids warnings when `options(warnPartialMatchArgs = TRUE)`.


mtmorgan:
@nturaga can you add a version bump, build & check in devel only, and if things look good merge?

nturaga:
The PR builds and checks fine against the devel version of Bioconductor. 

I'll merge it.

PR #133, opened 2020-12-24
DoparParam to handle also "orchestration" errors

### LINKED ISSUE #132: DoParam: Handle non-standard "orchestration" errors
# Issue

The error handling of the DoparParam backend makes assumptions about the format of the condition message that might not be valid.  [Specifically](https://github.com/Bioconductor/BiocParallel/blob/19e1e1386f3e616effcd1ba0479ebe2de0dd7875/R/DoparParam-class.R#L84-L93), it assumes it can parse the condition message to extract the "task index";

```r
    res <- tryCatch({
        foreach::foreach(X=X, .errorhandling=handle) %dopar% FUN(X, ...)
    }, error=function(e) {
        txt <- "'DoparParam()' does not support partial results"
        updt <- rep(list(.error_not_available(txt)), length(X))
        msg <- conditionMessage(e)
        i <- sub("task ([[:digit:]]+).*", "\\1", msg)
        updt[[as.integer(i)]] <- .error(msg)
        updt
    })
```

If the error message does not match the expected regular expression, then the inferred task index becomes `NA_integer_` and the subset assignment:

```r
msg <- conditionMessage(e)
i <- sub("task ([[:digit:]]+).*", "\\1", msg)
updt[[as.integer(i)]] <- .error(msg)
```

results in:

```r
Error in updt[[as.integer(i)]] <- .error(msg) : 
  attempt to select more than one element in integerOneIndex
In addition: Warning message:
In value[[3L]](cond) : NAs introduced by coercion
```

## Reproducible examples

This can happen if the error is produced by the foreach backend rather than the foreach expression.  Here's are a few reproducible examples, which rely on escalating backend warnings to errors.

### Example 1 - doSEQ

```r
library(BiocParallel)
register(DoparParam())
options(warn=1)
y <- bplapply(1L, identity)
## Warning: executing %dopar% sequentially: no parallel backend registered
```

```r
library(BiocParallel)
register(DoparParam())
options(warn=2) ## Escalate to an error
y <- bplapply(1L, identity)
## Error in value[[3L]](cond) : 
##   (converted from warning) NAs introduced by coercion
```

## Example 2 - Cluster worker crashes

```r
library(BiocParallel)
register(DoparParam())
cl <- parallel::makeCluster(1L)
doParallel::registerDoParallel(cl)
y <- bplapply(1L, function(x) quit("no"))
## Error in updt[[as.integer(i)]] <- .error(msg) : 
##   attempt to select less than one element in integerOneIndex
## In addition: Warning message:
## In value[[3L]](cond) : NAs introduced by coercion
```

### Example 3 - doRNG

```r
library(BiocParallel)
register(DoparParam())
doParallel::registerDoParallel(2L)
doRNG::registerDoRNG()
options(warn=1)
y <- bplapply(1L, function(x) { RNGkind("Super-Duper"); x })
Warning in (function (obj, ex)  :
  Foreach loop (doParallelMC) had changed the current RNG type: RNG was restored to same type, next state
```

```r
library(BiocParallel)
register(DoparParam())
doParallel::registerDoParallel(2L)
doRNG::registerDoRNG()
options(warn=2) ## Escalate to an error
y <- bplapply(1L, function(x) { RNGkind("Super-Duper"); x })
## Error in value[[3L]](cond) : 
##  (converted from warning) NAs introduced by coercion
```

### Example 4 - doFuture

```r
library(BiocParallel)
register(DoparParam())
doFuture::registerDoFuture()
options(future.rng.onMisuse = "warning")
y <- bplapply(1L, rnorm)
## Warning message:
## UNRELIABLE VALUE: One of the foreach() iterations (‘doFuture-1’) unexpectedly generated random numbers without declaring so. There is a risk that those random numbers are not statistically sound and the overall results might be invalid. To fix this, use '%dorng%' from the 'doRNG' package instead of '%dopar%'. This ensures that proper, parallel-safe random numbers are produced via the L'Ecuyer-CMRG method. To disable this check, set option 'future.rng.onMisuse' to "ignore".
```

```r
library(BiocParallel)
register(DoparParam())
doFuture::registerDoFuture()
options(future.rng.onMisuse = "error")  ## Escalate to an error
y <- bplapply(1L, rnorm)
Error in updt[[as.integer(i)]] <- .error(msg) : 
  attempt to select less than one element in integerOneIndex
In addition: Warning message:
In value[[3L]](cond) : NAs introduced by coercion
```

# Suggestion

Although it seems that the foreach authors, and other doNNN package developers, have adopted the de-facto standard to [relay errors as](https://github.com/RevolutionAnalytics/foreach/blob/731cff6e1acdd2ded4a0eceb87dc81e8bde1591e/R/do.R#L133-L136):

```r
  if (identical(obj$errorHandling, 'stop') && !is.null(errorValue)) {
    msg <- sprintf('task %d failed - "%s"', errorIndex,
                  conditionMessage(errorValue))
    stop(simpleError(msg, call=expr))
  }
```

I don't know if it has been documented or official be said that that must be the case.   Ideally, they would have formalized this as:

```r
  if (identical(obj$errorHandling, 'stop') && !is.null(errorValue)) {
    msg <- sprintf('task %d failed - "%s"', errorIndex,
                  conditionMessage(errorValue))
    ex <- simpleError(msg, call=expr)
    ex$errorIndex <- errorIndex
    ex$errorValue <- errorValue
    class(ex) <- c("foreachError", class(ex))
    stop(ex)
  }
```

or similar.  But, without that information, I understand that parsing the error message is the best thing that can be done.  However, I think one needs to be defensive about this assumption and have a fallback when it's not true.  Maybe something like:

```r
msg <- conditionMessage(e)
pattern <- "task ([[:digit:]]+).*"
if (!grepl(pattern, msg)) stop("An exceptional DoparParam + foreach() error occurred: ", msg)
i <- sub(pattern, "\\1", msg)
updt[[as.integer(i)]] <- .error(msg)
```

This would give the end-user a more informative error message.

I'll try to do a PR next.

PS. In the future ecosystem, I call these types of errors "orchestration errors" - they're not run-time errors that occur from evaluating the indented expression or function call but rather error that occurs to orcestrate the map-reduce or parallel processing.

HenrikBengtsson:
Fixes #132

### DIFF
--- a/R/DoparParam-class.R
+++ b/R/DoparParam-class.R
@@ -87,7 +87,10 @@ setMethod("bplapply", c("ANY", "DoparParam"),
         txt <- "'DoparParam()' does not support partial results"
         updt <- rep(list(.error_not_available(txt)), length(X))
         msg <- conditionMessage(e)
-        i <- sub("task ([[:digit:]]+).*", "\\1", msg)
+        pattern <- "task ([[:digit:]]+).*"
+        if (!grepl(pattern, msg))
+            stop("An exceptional DoparParam + foreach() error occurred: ", msg)
+        i <- sub(pattern, "\\1", msg)
         updt[[as.integer(i)]] <- .error(msg)
         updt
     })


HenrikBengtsson:
Och m

Issue #122, opened 2020-10-02
MulticoreParam RNGseed does not work


mtmorgan:
Expecting the same result from each evaluation
```
p = MulticoreParam(2, RNGseed = 123)
bplapply(1:2, function(...) rnorm(1), BPPARAM = p)
bplapply(1:2, function(...) rnorm(1), BPPARAM = p)
```
but get, e.g.,
```
> p = MulticoreParam(2, RNGseed = 123)
> str(bplapply(1:2, function(...) rnorm(1), BPPARAM = p))
List of 2
 $ : num 0.00494
 $ : num 0.897
> str(bplapply(1:2, function(...) rnorm(1), BPPARAM = p))
List of 2
 $ : num 1.81
 $ : num -0.498
```

DarwinAwardWinner:
Even recreating the param with the same seed doesn't work:

```r
p1 = MulticoreParam(2, RNGseed = 123)
bplapply(1:2, function(...) rnorm(1), BPPARAM = p1)
p2 = MulticoreParam(2, RNGseed = 123)
bplapply(1:2, function(...) rnorm(1), BPPARAM = p2)
```

Result:

```
> p1 = MulticoreParam(2, RNGseed = 123)
> bplapply(1:2, function(...) rnorm(1), BPPARAM = p1)
[[1]]
[1] 2.60496

[[2]]
[1] 0.0878279

> p2 = MulticoreParam(2, RNGseed = 123)
> bplapply(1:2, function(...) rnorm(1), BPPARAM = p2)
[[1]]
[1] 0.648661

[[2]]
[1] 0.1057389
```

mtmorgan:
This is just with the 'transient' MulticoreParam implementation, where the workers have not been forked before the call to `bplapply()`. So for instance the following works

```
> bpstart(p)
> str(bplapply(1:5, function(...) rnorm(1), BPPARAM = p))
List of 5
 $ : num -0.969
 $ : num 0.706
 $ : num -0.409
 $ : num 0.891
 $ : num -0.865
> bpstop(p)
> bpstart(p)
> str(bplapply(1:5, function(...) rnorm(1), BPPARAM = p))
List of 5
 $ : num -0.969
 $ : num 0.706
 $ : num -0.409
 $ : num 0.891
 $ : num -0.865
> bpstop(p)
```

I understand the problem and will think about a fix...

wmacnair:
[...]

Issue #115, opened 2020-04-21
MulticoreParam and random port setting advances the random stream


LTLA:
It seems that MulticoreParam() still modifies the random seed:

```r
set.seed(10)
runif(1)
## [1] 0.5074782

set.seed(10)
MulticoreParam(1)
runif(1)
## [1] 0.3067685
```

This only occurs in BioC-devel (and release, presumably, but I don't work there). It does not occur in the GitHub version, courtesy of the setting and unsetting of the internal RNG stream.

https://github.com/Bioconductor/BiocParallel/blob/15dc9f6a4c8d7d11a120831412824cba0ac82f69/R/SnowParam-class.R#L26-L27

Presumably there was a failure to propagate this to the build system for some reason. Downloading the BiocParallel tarball shows version 1.21.2 but without the `internal_rng_stream.R` file... weird.

<details>
<summary>Session information</summary>

```
R Under development (unstable) (2020-03-23 r78035)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 18.04.4 LTS

Matrix products: default
BLAS:   /home/luna/Software/R/trunk/lib/libRblas.so
LAPACK: /home/luna/Software/R/trunk/lib/libRlapack.so

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] BiocParallel_1.21.2

loaded via a namespace (and not attached):
[1] compiler_4.0.0 parallel_4.0.0
```
</details>

Issue #110, opened 2019-12-22
BiocParallel and random seeds, redux


LTLA:
Consider the following code:

```r
library(BiocParallel)
BPPARAM <- MulticoreParam(5)

set.seed(100)
runif(10)
## [1] 0.30776611 0.25767250 0.55232243 0.05638315 0.46854928 0.48377074
## [7] 0.81240262 0.37032054 0.54655860 0.17026205

set.seed(100)
bplapply(1:3, seq_len, BPPARAM=BPPARAM)
runif(10)
## [1] 0.05638315 0.46854928 0.48377074 0.81240262 0.37032054 0.54655860
## [7] 0.17026205 0.62499648 0.88216552 0.28035384
```

I appreciate that some randomization is necessary when setting up the backend but this behavior is unexpected. I would have thought that the perturbation of the seed would have been done in the constructor, so any `set.seed()` after that would guarantee consistent results. In this case, it seems that every job is hitting the PRNG exactly once as seen by the shift in the random numbers above; replacing with `SnowParam()` eliminates the difference.

For some context, I have functions that use `bplapply`, etc. and perform randomization **after** the `bplapply` call; as a result of the behavior above, the output of these functions is no longer agnostic to the choice of `BPPARAM`. This seems undesirable to me.

<details>
<summary>Session information</summary>

```
R Under development (unstable) (2019-10-31 r77342)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 18.04.3 LTS

Matrix products: default
BLAS:   /home/luna/Software/R/trunk/lib/libRblas.so
LAPACK: /home/luna/Software/R/trunk/lib/libRlapack.so

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] BiocParallel_1.21.1

loaded via a namespace (and not attached):
[...]

PR #140, opened 2021-04-27
Demonstration of job-specific seeds

LTLA:
This is a proof of concept for job-specific seeds. The idea is that a seed (conflated with stream for the L'Ecuyer RNG) is defined for each job, usually in the serial section ; the relevant seeds are distributed along with the other job parameters to the workers; and the seed is set to the specified value prior to each call to `FUN`. This ensures that, no matter how the jobs are split up among the workers, the same result is obtained. 

The same cannot be said for the current set-up. Partly because the seed is set once per worker, such that changes in the number of workers will yield different results; but mostly because of #122, which means that the seed isn't really respected anyway for the most common use case of `MulticoreParam()`.

The current PR has known gaps - doesn't work for `SerialParam()`, for example - but these can be easily fixed.

### DIFF
--- a/R/ErrorHandling.R
+++ b/R/ErrorHandling.R
@@ -66,9 +66,12 @@ bptry <- function(expr, ..., bplist_error, bperror)
         else e
     }
 
-    function(...) {
+    function(..., .BiocParallel.seed) {
         setTimeLimit(timeout, timeout, TRUE)
         on.exit(setTimeLimit(Inf, Inf, FALSE))
+
+        RNGkind("L'Ecuyer-CMRG")
+        assign(".Random.seed", .BiocParallel.seed, envir = .GlobalEnv)
 
         if (exportglobals)
             base::options(global_options)
--- a/R/ErrorHandling.R
+++ b/R/ErrorHandling.R
@@ -70,7 +70,8 @@ bptry <- function(expr, ..., bplist_error, bperror)
         setTimeLimit(timeout, timeout, TRUE)
         on.exit(setTimeLimit(Inf, Inf, FALSE))
 
-        RNGkind("L'Ecuyer-CMRG")
+        old <- RNGkind("L'Ecuyer-CMRG")
+        on.exit(RNGkind(old[1]))
         assign(".Random.seed", .BiocParallel.seed, envir = .GlobalEnv)
 
         if (exportglobals)
--- a/R/bplapply-methods.R
+++ b/R/bplapply-methods.R
@@ -73,12 +73,16 @@ setMethod("bplapply", c("ANY", "list"),
         timeout=bptimeout(BPPARAM), exportglobals=bpexportglobals(BPPARAM)
     )
 
+    ## manufacture seeds
+    seeds <- .define_seeds(bpRNGseed(BPPARAM), length(X))
+
     ## split into tasks
     X <- .splitX(X, bpnworkers(BPPARAM), bptasks(BPPARAM))
-    ARGFUN <- function(i) c(list(X=X[[i]]), list(FUN=FUN), list(...))
+    seeds <- .splitX(seeds, bpnworkers(BPPARAM), bptasks(BPPARAM))
+    ARGFUN <- function(i) c(list(X[[i]], .BiocParallel.seed=seeds[[i]]), list(FUN=FUN), list(MoreArgs=list(...), SIMPLIFY=FALSE))
 
     cls <- structure(list(), class="lapply")
-    res <- bploop(cls, X, lapply, ARGFUN, BPPARAM)
+    res <- bploop(cls, X, mapply, ARGFUN, BPPARAM)
 
     if (!is.null(res)) {
         res <- do.call(unlist, list(res, recursive=FALSE))
--- a/R/bpstart-methods.R
+++ b/R/bpstart-methods.R
@@ -98,7 +98,8 @@ setMethod("bpstart", "missing",
         if (exists(".Random.seed", envir = .GlobalEnv, inherits = FALSE)) {
             get(".Random.seed", envir = .GlobalEnv, inherits = FALSE)
         } else NULL
-    RNGkind("L'Ecuyer-CMRG")
+    old <- RNGkind("L'Ecuyer-CMRG")
+    on.exit(RNGkind(old[1]))
     if (!is.null(init_seed))
         set.seed(init_seed)
     seeds <- vector("list", n)
--- a/R/bpstart-methods.R
+++ b/R/bpstart-methods.R
@@ -73,25 +73,8 @@ setMethod("bpstart", "missing",
 {
     cluster <- bpbackend(x)
     rng_seed <- bpRNGseed(x)
-
-    ##
-    ## from parallel::clusterSetRNGStream
-    ##
-    oldseed <-
-        if (exists(".Random.seed", envir = .GlobalEnv, inherits = FALSE)) {
-            get(".Random.seed", envir = .GlobalEnv, inherits = FALSE)
-        } else NULL
-    RNGkind("L'Ecuyer-CMRG")
-    if (!is.null(rng_seed))
-        set.seed(rng_seed)
     nc <- length(cluster)
-    seeds <- vector("list", nc)
-    seeds[[1L]] <- .Random.seed
-    for (i in seq_len(nc - 1L))
-        seeds[[i + 1L]] <- nextRNGStream(seeds[[i]])
-    if (!is.null(oldseed)) {
-        assign(".Random.seed", oldseed, envir = .GlobalEnv)
-    } else rm(.Random.seed, envir = .GlobalEnv)
+    seeds <- .define_seeds(rng_seed, nc)
 
     for (i in seq_along(cluster)) {
         expr <- substitute(
@@ -105,6 +88,28 @@ setMethod("bpstart", "missing",
 
     .bpstart_error_handler(x, response, "set_rng_seed")
     invisible(x)
+}
+
+.define_seeds <- function(init_seed, n) {
+    ##
+    ## from parallel::clusterSetRNGStream
+    ##
+    oldseed <-
+        if (exists(".Random.seed", envir = .GlobalEnv, inherits = FALSE)) {
+            get(".Random.seed", envir = .GlobalEnv, inherits = FALSE)
+        } else NULL
+    RNGkind("L'Ecuyer-CMRG")
+    if (!is.null(init_seed))
+        set.seed(init_seed)
+    seeds <- vector("list", n)
+    seeds[[1L]] <- .Random.seed
+    for (i in seq_len(n - 1L))
+        seeds[[i + 1L]] <- nextRNGStream(seeds[[i]])
+    if (!is.null(oldseed)) {
+        assign(".Random.seed", oldseed, envir = .GlobalEnv)
+    } else rm(.Random.seed, envir = .GlobalEnv)
+
+    seeds
 }
 
 .bpstart_set_finalizer <-


DarwinAwardWinner:
Can this approach be extended to `bpiterate`, in which the number of tasks is not known ahead of time?

LTLA:
I think so, in principle. The idea would be to run `nextRNGStream()` to generate a new seed for each job as it spawns. Then you'll always get the same result provided the jobs always spawn in the same order.

LTLA:
This system is now in use in **SingleR**'s `aggregateReference()` function.

mtmorgan:
This will be implemented shortly after the next release, in the devel branch

mtmorgan:
I've finally implemented a solution on the [rng branch][], and would welcome any feedback.

[rng branch]: https://github.com/Bioconductor/BiocParallel/tree/rng

```
library(BiocParallel)

FUN <- function(i) rnorm(1)

## no seed set -- independent results
p0 <- MulticoreParam(3)
result <- unlist(bplapply(1:11, FUN, BPPARAM = p0))
!any(result == unlist(bplapply(1:11, FUN, BPPARAM = p0)))

p0 <- SnowParam(3)
result <- unlist(bplapply(1:11, FUN, BPPARAM = p0))
!any(result == unlist(bplapply(1:11, FUN, BPPARAM = p0)))

## seed set -- same results across worker, thread, and *Param
p1 <- MulticoreParam(3, RNGseed = 123)
p2 <- MulticoreParam(5, RNGseed = 123)
p3 <- MulticoreParam(5, 7, RNGseed = 123)
p4 <- SnowParam(3, RNGseed = 123)
p5 <- SerialParam(RNGseed = 123)

result <- unlist(bplapply(1:11, FUN, BPPARAM = p1))
identical(result, unlist(bplapply(1:11, FUN, BPPARAM = p1)))
identical(result, unlist(bplapply(1:11, FUN, BPPARAM = p2)))
identical(result, unlist(bplapply(1:11, FUN, BPPARAM = p3)))
identical(result, unlist(bplapply(1:11, FUN, BPPARAM = p4)))
identical(result, unlist(bplapply(1:11, FUN, BPPARAM = p5)))

## bplapply / bpiterate same results
ITER_factory <- function() {
    x <- 1:11
    i <- 0L
    function() {
        i <<- i + 1L
        if (i > length(x))
            return(NULL)
        x[[i]]
    }
}

unlist(bpiterate(ITER_factory(), FUN, BPPARAM = p1))
identical(result, unlist(bpiterate(ITER_factory(), FUN, BPPARAM = p1)))
identical(result, unlist(bpiterate(ITER_factory(), FUN, BPPARAM = p2)))
identical(result, unlist(bpiterate(ITER_factory(), FUN, BPPARAM = p3)))
identical(result, unlist(bpiterate(ITER_factory(), FUN, BPPARAM = p4)))
identical(result, unlist(bpiterate(ITER_factory(), FUN, BPPARAM = p5)))
```

LTLA:
This would work best if, when `RNGseed=NULL`, an appropriate value is drawn from R's global session seed. This would allow the following to yield reproducible results:

```
p1 <- MulticoreParam(3)
set.seed(100); result <- unlist(bplapply(1:11, FUN, BPPARAM = p1))
set.seed(100); result2 <- unlist(bplapply(1:11, FUN, BPPARAM = p1))
```

Otherwise, users would have to learn about a new seed setting mechanism, and developers would have to inspect the incoming `BPPARAM` object to check whether a **BiocParallel** seed is properly set.

mtmorgan:
I'll probably do this.

What do you mean 'developers would have to inspect...'? Shouldn't packages just respect whatever the user does?

One thing about the current situation is that the global random number seed is never incremented (for bplapply / bpiterate, Serial, Snow, or MulticoreParam). Whereas with the behavior above it would sometimes be iterated -- so a loss of consistency.

A second concern is when the RNGseed is 'captured'. It could be when `MulticoreParam()` is created, but then the example above would not result in reproducibility. The above assumes that the seed is captured on `bpstart()` (which is implicit on passing a stopped `p1` to `bplapply()`).  The following would result in different results
1 & 2.

```
set.seed(100); p1 <- bpstart(MulticoreParam(3))
set.seed(100); result <- unlist(bplapply(1:11, FUN, BPPARAM = p1))
set.seed(100); result2 <- unlist(bplapply(1:11, FUN, BPPARAM = p1))
```

These situations quickly lead the user into the details of the implementation, and I wonder whether it is better to just adopt behaviors up front that are consistent, but that require the user to know what they are doing...?

DarwinAwardWinner:
Worth nothing, the future package solves all these issues and provides the ability to do fully reproducible RNG. That is, when reproducible RNG is requested, the same code will always produce identical results, regardless of the backend used, number of parallel jobs, etc. (Apparently reproducible RNG must be explicitly requested because generating parallel streams of random numbers is expensive.) Here's the best description I've found of how future does it: https://www.r-bloggers.com/2020/09/future-1-19-1-making-sure-proper-random-numbers-are-produced-in-parallel-processing/

Note that as far as I know, the future package *doesn't* guarantee that, e.g. `future_lapply` will return the same as `lapply`, just that any call to `future_lapply` that requests reproducible RNG will give the same output given the same initial RNG state.

mtmorgan:
Commit 23c6864c788abe14f2ade0663c6208548000e6b6 uses the global stream when `RNGseed = NULL`. The global stream is incremented exactly once whenever `bpstart()` is called, implicitly or explicitly, regardless of `RNGseed` value. `set.seed(123); result1 <- bplapply(1:3, runif, BPPARAM = SerialParam())` and `set.seed(124); result2 <- bbplappyl(1:3, runif, BPPARAM = SerialParam(RNGseed = 123))` produce identical `result1` and `result2` -- when an `RNGseed` is provided, the global stream is set before creating the element stream (and reset and incremented after).

LTLA:
> What do you mean 'developers would have to inspect...'? Shouldn't packages just respect whatever the user does?

If `BPPARAM` didn't respect the global stream and `RNGseed` was not set, then a developer wishing to achieve reproducibility 
on parallel execution would have to take matters into their own hands; hence the need to inspect whether `RNGseed = NULL`.

LTLA:
> The global stream is incremented exactly once whenever `bpstart()` is called, implicitly or explicitly, regardless of `RNGseed` value.

I just realized that this breaks existing code that expects **BiocParallel** to not touch the stream. This is the cause for a number of current build failures, e.g., **BiocSingular**, **scran**, **scater**. where tests assume that the stream is not altered. This is but a symptom as the change in the stream will (unnecessarily) change user-visible results in the next release.

Why not draw the seed inside `bplapply()` itself? This is what was originally proposed in the PR; it avoids any questions and ambiguity around the interaction between the seed and the creation of the `BPPARAM` object. Just define the seed inside `bplapply()` and distribute it with the function to be executed. Additionally provide an option in `bplapply()` to indicate whether seed passing is desired - the default should be no, so as to preserve consistency with previous results.

mtmorgan:
@LTLA this behavior came from your earlier comment, that when `RNGseed = NULL` the global random number stream is used. But since
```
set.seed(123)
result1 = lapply(1:3, runif)
result2 = lapply(1:3, runif)
stopifnot(!identical(result1, result2))
```
it seems like
```
set.seed(123)
result1 = bplapply(1:3, runif)
result2 = bplapply(1:3, runif)
stopifnot(!identical(result1, result2))
```
and hence that the global random number stream is incremented (or somehow `bplapply()` were to know that the globbal random number seed was set to `123` on first call, but not on second, but this seems like a hack...) on calls to `bplapply()` (actually, `bpstart()`).

So it sounds like you'd rather violate this expectation, and have
```
set.seed(123)
result1 = bplapply(1:3, runif)
result2 = bplapply(1:3, runif)
stopifnot(identical(result1, result2))
```

LTLA:
After some thought, I think that the fundamental problem is that `bplapply()` and friends do not know whether the supplied `FUN()` will involve randomization. 

If it _doesn't_ involve randomization, then `bplapply()` definitely shouldn't modify the global stream. 

If it _does_ involve randomization, then `bplapply()` _should_ modify the global stream reproducibly, consistent with any other function that does randomization. The problem being, of course, that any modification of the global stream must occur in the serial section to create the job-specific seeds before `FUN()` itself is actually called.

I think the solution is to allow developers to somehow indicate to `bplapply()` that their `FUN` will involve randomization, in which case the job-specific seeds are initialized; otherwise, the global stream is left untouched. One option is to allow developers to wrap the supplied `BPPARAM` in a `RNGSeedWrapperParam()` wrapper that is then supplied to `BPPARAM=`; the various **BiocParallel** methods can then recognize the wrapper, initialize the job-specific seeds, and extract the wrapped `BPPARAM` for actual dispatch. This approach avoids having to add more arguments to `bplapply()` (and thus potentially conflicting with the `...` arguments) and avoids the addition of more slots to existing `BiocParallelParam` objects.

Another option is to overload on the `RNGseed` slot; if `bplapply()` sees it being set, it creates job-specific seeds, otherwise it does nothing. I don't like this idea as much, because users might already be pre-emptively setting `RNGseed` "just in case" and it would require an active effort on behalf of the developer to unset it to avoid fiddling with the global stream.

HenrikBengtsson:
> If it doesn't involve randomization, then bplapply() definitely shouldn't modify the global stream.
> 
> If it does involve randomization, then bplapply() should modify the global stream reproducibly, consistent with any other function that does randomization. The problem being, of course, that any modification of the global stream must occur in the serial section to create the job-specific seeds before FUN() itself is actually called.
> 
> I think the solution is to allow developers to somehow indicate to bplapply() that their FUN will involve randomization, in which case the job-specific seeds are initialized; otherwise, the global stream is left untouched.

This is the design used by **future.apply** and **furrr**.  It's only the developer who can decide whether or not RNG is needed.  This is specified via argument `future.seed = TRUE`.  It's also cheap to check if the RNG was used by mistake at the end.  This is discussed in <https://www.jottr.org/2020/09/22/push-for-statistical-sound-rng/>.

mtmorgan:
I guess code will need to be changed regardless? So that shouldn't be a driving consideration.

My current feeling is to adopt the following behavior

- `RNGseed = NULL` (default) -- use an internal stream to set a random seed. Iterate the internal stream. Do not iterate the global stream.
- `RNGseed = integer(1)` -- a user-supplied seed used for reproducibility. Do not iterate the internal stream. Do not iterate the global stream.
- `RNGseed = NA` -- use the global stream. Do not iterate the internal stream. Iterate (by 1) the global stream for each call to `bpstart()`.

I cannot tell if this addresses Aaron's earlier comment about the developer 'needing to inspect' the value of `RNGseed` in https://github.com/Bioconductor/BiocParallel/pull/140#issuecomment-916675434 . I guess I envisioned

```
user_facing_fun <- function(..., BPPARAM = SnowParam(RNGseed = 123)) {
    ## internal implementation
}
```

The user gets to override the seed, but `user_facing_fun()` is reproducible when invoked 'out of the box'.

LTLA:
I don't like the proposed `RNGseed` solution. This asks too much of the user with respect to understanding the interaction between **BiocParallel**'s internal stream, the global stream, and whether or not the `FUN` even does randomization. We should be making this all transparent; the end user should not have to think of anything except calling `set.seed()`. Developers should be the only ones worrying about how the interaction occurs and making sure that their functions behave as expected. 

If you go with this `RNGseed` solution, I will forcibly set `RNGseed = NA` in all my **BiocParallel** functions involving randomization to force use of the global stream. More importantly, I will now have to set `RNGseed = 0` in all my other **BiocParallel** functions to prevent any use of the global stream, regardless of what the user asks. Otherwise, anyone who sets `RNGseed = NA` will now observe weird changes in the global stream in functions that have no randomization.

mtmorgan:
Ok, next proposal. 

- `RNGseed = NULL`. (default) Each call to `bplapply()` / `bpiterate()` seeds random number streams from the current global stream.

  Only if `FUN()` uses the random number stream, Increment the stream either (1) by `n`, the number of times random numbers were actually generated, so that the stream is in the same state 'as if' `lapply()` had been called. From the perspective of the calling calling code `bplapply()` and `lapply()` have the same consequences for the global stream. But of course the code in `FUN` would see different random numbers. Or (2) increment the stream by 1, because only one call to the global stream is used to seed the workers. Figuring out `n` would involve counting how many calls to, e.g., `runif(1)` it takes to move the initial seed assigned to a worker to the final value seen at the end of the worker task. This would break (perhaps badly) if user code did things like `set.seed()` inside `FUN`. So option 2 seems easy and robust but not completely satisfactory.
- `RNGseed = integer(1)`. Use this number to seed the random number streams in `bpstart()` (implictly `bplapply()` / `bpiterate()`).

LTLA:
That seems pretty complicated to implement and might add some substantial overhead for short-running `FUN()`. Why not let the developer decide whether or not a seed should be set? That would be simple and explicit, no need for guesswork.

mtmorgan:
Incrementing the global stream by 1 only if the random number stream is used is easy and fast to implement -- is the seed at the end of FUN the same as it was at the beginning, so a single call to `identical()`.

> Why not let the developer decide whether or not a seed should be set?

Isn't that what `RNGseed = 0` does? `bplapply()` does whatever behind the scenes (who cares that it sets up an appropriate random number environment that is not used?), but the global random number generator is not incremented.

LTLA:
Well, the seeds are hardly small vectors; mine is 626 integers long.

The main complication comes from the interaction with existing package/user code making implicit assumptions about the state of the seed after `bplapply()`. At the very least, the new default will yield a different state for the global stream than what was there before, for developers who were already using a randomized `FUN()` in `bplapply()`. My point is that, whatever mechanism we use, it should be an opt-in system to use the new behavior. Give us a chance to migrate rather than "surprise, here it is".

mtmorgan:
The correct RNG is L'Ecuyer, where the seed size is six integers. With the default mode for distributing a job, where the length of X is split into a number of vectors equal to the number of `n` workers (e.g., 12), only an additional `n x 6` integers are serialized, and only an additional `n` scalar logicals returned.

It's not possible (ok, everything is possible...) to recover the old behavior, because the old behavior was inconsistent, even within a `*Param()`.

```
> set.seed(123); res <- bplapply(1:3, runif, BPPARAM = SnowParam(3)); runif(1)
[1] 0.7883051
> set.seed(123); res <- bplapply(1:3, runif, BPPARAM = SnowParam(3)); runif(1)
[1] 0.2875775
```
!! (ok, obviously not anything to be proud of, or easy to retain with a simple 'do things the olde way' option). I guess I'm trying to take advantage of the 'devel' branch to introduce breaking changes. And user who must have identical behavior also has a mechanism -- continue using the current 'release' branch.

LTLA:
Well, if I'm calling `set.seed()` inside my `FUN`, then the new default behavior will change the state of the global stream, because `bplapply()` is now jumping ahead by 1 when it detects that the seed is changed. This causes changes to the output of downstream randomization steps - and subsequently, the analysis results. This would be a bitter pill to swallow when I put `set.seed()` inside `FUN` to obtain per-job reproducibility in the first place.

Why not just tweak the scheme proposed in an earlier comment?

- `RNGseed = NULL`, which is the current default, is not changed. Nothing is done with respect to bumping the stream.
- `RNGseed=<some integer>` does whatever it was doing. I don't use this so I don't particularly care what changes here.
- `RNGseed=NA` does the new behavior and developers can opt into this.

Jiefei-Wang:
> Well, if I'm calling `set.seed()` inside my `FUN`, then the new default behavior will change the state of the global stream after my function is called, because `bplapply()` is now jumping ahead by 1 when it detects that the seed is changed. This causes changes to the output of downstream randomization steps - and subsequently, the analysis results. This would be a bitter pill to swallow when I put `set.seed()` inside `FUN` to obtain per-job reproducibility in the first place.

I'll argue that this is the wrong way to write your code, it does not matter if you know the seed implemented in `BiocParallel`. 

If the user does not know how the seed is implemented, at least he knows this `set.seed` is run in the worker. It does not make sense to assume `set.seed` is able to set the global seed, and so give you a reproducible result after `bplapply`(but he can have a reproducible result if `set.seed` is called before `bplapply`, and the code after `bplapply` is also reproducible since the iteration of the global seed is fixed)

If the user knows the implementation of the seed, then why does he still have this expectation?

If we really want to take care of this use case, we can use the seed returned by the last task result from the worker and set the global seed accordingly. Therefore
```
res <- bplapply(1:3, function(x) {
    set.seed(1)
    runif(1)
})
runif(1)
```
the `runif(1)` in the last line will always use the seed from `FUN(3)` in `bplapply`, which can be reproducible.

mtmorgan:
Using `set.seed()` inside FUN is wrong, at least in principle, because it undermines the independence of the streams. I think?

The behavior in the current release branch ignores the state of the global random number generator, so `set.seed()` is irrelevant. Earlier @LTLA said (https://github.com/Bioconductor/BiocParallel/pull/140#issuecomment-908068391) it would be a good idea to pay attention to the the global stream so the user could just do `set.seed(123); bplapply(1:3, runif)` and get reproducible results. Basically this sounds like a very good idea to me.

As noted in https://github.com/Bioconductor/BiocParallel/pull/140#issuecomment-920267227 using the global stream to initialize the random number stream necessitates (at least when random numbers are actually used in `FUN`) incrementing the global stream. Otherwise `res1 <- bplapply(1:3, runif); res2 <- bplapply(1:3, runif)` would produce `identical(res1, res2)`.

I initially opted for 'alway increment the stream by 1' instead of 'increment the stream by 1 if random numbers are used in `FUN`', so the user wouldn't have to understand whether and when `FUN` used random numbers, directly or indirectly. This reduces cognitive load on users and package developers (Henrik's blog post recognizes the challenges of identifying when and where code uses the random number stream). I think the consistency of 'always increment by 1' actually enhances reproducibility. For instance, if the rule is 'only increment the global stream if the function uses random numbers', then one could imagine that for some inputs `FUN` follows a branch that uses random numbers, and for other inputs does not. The state of the global stream then depends on which branch of `FUN` was executed, so one cannot in general rely on the global stream not being incremented for reproducibility in the larger code base. Maybe an interesting illustration of this is `sample(n)`. It's not obvious how many calls to a random number generator are needed to create a permutation of size `n` -- it seems like it could be just 1, or maybe `n - 1` (once you've created a permutation of `n - 1` numbers, the full permutation is known)?. Actually, `sample(n)` increments the random number stream by some more complicated function of `n`, and actually doesn't increment the stream with `sample(0)` or `sample(1)`. It seems like it would be much easier to reason about reproducibility if `sample(n)` incremented the global stream by 1, always.

I shied away from 'increment the stream only if the user or developer says so', which is the approach in Henrik's packages, because I think the code should do the right thing, rather than the wrong thing followed by an admonition to take on the cognitive load / reduced reproducibility comments of the previous paragraph.

I'm not sure where this leaves us... the closest thing to a non-breaking change would be to ignore `set.seed()`. But this could still break existing code, so why not aim for something better?

HenrikBengtsson:
> Using set.seed() inside FUN is wrong, at least in principle, because it undermines the independence of the streams. I think?

I'm in this camp too. I think `set.seed()` is something that belongs to the end-user at the top of their script, and basically only once, if at all.  Using `set.seed()` in functions or packages is probably not correct and most like used due to some ad-hoc I-think-this-should-be-good-enough-for-now algorithms for parallel RNGs, e.g. `for (ii in 1:100) { set.seed(ii); ... }` and versions thereof.  (On a side node, search CRAN/GitHub and you'll find lots of functions that calls `set.seed()` internally. The really scary ones are the ones that call `set.seed(<constant>)` - that can really mess up your analysis without you noticing.)

> ... Actually, sample(n) increments the random number stream by some more complicated function of n, and actually doesn't increment the stream with sample(0) or sample(1). It seems like it would be much easier to reason about reproducibility if sample(n) incremented the global stream by 1, always.

If it wouldn't be for these type of conditional use of RNG, and functions either forward the RNG state or not, then one could imagine extending the R API so that functions had to declare whether or not they forward the RNG state, e.g. `attr(rnorm, "usesRNG") <- TRUE`.  With that we could use static-code inspection to infer if parallel RNG is needed.  Without that, I think the best is for the developer to declare the need for the RNG.

HenrikBengtsson:
> I shied away from 'increment the stream only if the user or developer says so', which is the approach in Henrik's packages, because I think the code should do the right thing, rather than the wrong thing followed by an admonition to take on the cognitive load / reduced reproducibility comments of the previous paragraph.

The problem is that producing parallel RNG streams is really expensive, especially if you do it for every single element ("perfect numerical reproducibility").  I went back and forth what the default should be.  The alternative could have been to produce RNG streams by default, and then check if they were really needed afterward, and give a warning that it should be safe to disable it.  I decided for an active decision by the developer for now in the hope to bring awareness.  Since the warning is instant and bold, it's likely that the user/developer investigates it. It's clear that several package developers took actions after these warnings were introduced.  It's possible that I switch the default at a later point.

mtmorgan:
I'm not sure that I understand the expense, @HenrikBengtsson -- it takes about 1/2 second to generate 1 million independent streams
```
> RNGkind("L'Ecuyer-CMRG"); set.seed(123); seed <- .Random.seed
> system.time(for (i in 1:1000000) seed <- nextRNGStream(seed))
   user  system elapsed
  0.580   0.002   0.582
```
And as mentioned in https://github.com/Bioconductor/BiocParallel/pull/140#issuecomment-921367827 the (short integer vector) seeds only need to be serialized a few times for most applications.

I'll also say that when one talks about starting new processes with SNOW-style parallelism, the overhead of start-up and of serialized movement of data to workers means that the it only really pays to use parallel evaluation when there is some substantial work to be done in `FUN`. Hence the added cost of generating a seed and transmitting it will be a small fraction of the space and time requirements of an iteration of `FUN` -- R-level solutions are really targeting coarse-grained parallelism; for fine scale problems one might look to, e.g., parallel BLAS or implementations in C using Intel Threaded Building Blocks.

HenrikBengtsson:
My memory is bad, I'm actually generating also [substreams](https://github.com/HenrikBengtsson/future.apply/blob/c5186636398d327beb41104c80d99aa67b2479df/R/rng.R#L175-L186) in this process;

```r
  seeds <- vector("list", length = count)
  for (ii in seq_len(count)) {
    ## RNG substream seed used when calling FUN() for element(s) 'ii':
    ## This way each future can in turn generate further seeds, also
    ## recursively, with minimal risk of generating the same seeds as
    ## another future. This should make it safe to recursively call
    ## future_*apply(). /HB 2017-01-11
    seeds[[ii]] <- nextRNGSubStream(.seed)
    
    ## Main random seed for next iteration (= ii + 1)
    .seed <- nextRNGStream(.seed)
  }
```

The gist of the problem is that otherwise you risk having the same RNG stream in _nested_ calls, e.g.

```r
library(parallel)

make_seeds <- function(n) {
  seed <- .Random.seed
  seeds <- vector("list", length = n)
  for (ii in seq_len(n)) seeds[[ii]] <- seed <- nextRNGStream(seed)
  seeds
}

RNGkind("L'Ecuyer-CMRG")
set.seed(42)

## RNG seeds for three parallel workers
seeds <- make_seeds(3)

## On worker 1, generate seeds for two nested workers
.Random.seed <- seeds[[1]]
seeds_nested <- make_seeds(2)

## Turns out that nested worker (1,1), gets RNG seed:
seeds_nested[[1]]
#> [1]       10407  -358246211 -1411857254  -488221791  1069515918 -1720689715
#> [7]  1070199760

## ... which is the same as the RNG seed for worker 2 (BAD):
seeds[[2]]
#> [1]       10407  -358246211 -1411857254  -488221791  1069515918 -1720689715
#> [7]  1070199760
```


My somewhat ad-hoc workaround for this problem is to do:

```r
library(parallel)

make_seeds_2 <- function(n) {
  seed <- .Random.seed
  seeds <- vector("list", length = n)
  for (ii in seq_len(n)) {
    seeds[[ii]] <- nextRNGSubStream(seed)
    seed <- nextRNGStream(seed)
  }
  seeds
}
```

Admitadely, I don't recall how deeply I investigated the literature whether or not this is the proper way to do it.

LTLA:
> Using `set.seed()` inside FUN is wrong, at least in principle, because it undermines the independence of the streams. I think?

In my specific `SingleR` package code, it doesn't matter, because I manually set the seed again outside of the `bplapply()` loop. However, I can't say the same for others - for example, a user might be expecting this to be reproducible:

```r
library(BiocParallel)
set.seed(123); bplapply(1:5, function(x) { set.seed(x); runif(1) }); runif(1) # 0.2875775
set.seed(123); bplapply(1:5, function(x) { set.seed(x); runif(1) }); runif(1) # 0.2875775
```

Sure, this might not work robustly with other `BPPARAM` choices, but it's reproducible enough on a user's machine. So users might not expect this script to be particularly buggy. In fact, empirical testing indicates that it gives the same result for `nworkers > 1` on `MulticoreParam`, so people may even consider this reproducible across different MacOS/Linux machines.

The proposed change to the default will violate these expectations of reproducibility. You might say that these expectations are misplaced, but well, the fact of the matter is that people won't be able to reproduce their results. Now, it's debatable how much consistency can be expected across BioC releases, but I would suggest that "as much as possible" should be the goal here.

Which leads to my "opt-in" suggestion above, which gives the users/developers a chance to manage any change in results. This also aligns with what I believe to be a fundamental principle of **BiocParallel**'s design - that **BiocParallel** functions will never (unless prompted) alter the global stream. I don't know if this is an explicit policy but I've always complained about it when it happens (#110, #115, probably a few slack conversations, can't remember) so it's a _de facto_ expectation for me.

If you MUST change the default, then we could at least go through the same warning cycle that we have for all other non-transparent changes. That is, detect if there's a change in the seed and raise a warning to suggest users use `RNGseed=NA`, but don't do anything until the next release when that becomes the new default.

Jiefei-Wang:
I would like to go back a little bit and discuss a less technical question: `What assumptions and expectations the end-user will have when using the parallel package? `

If one asks me to pick up just a single principle for using the parallel package, I will say
```
The task sent from the master will be executed in the worker
```
Even though it looks like another way of describing parallel computing, but this assumption is actually important. It implies that there should be no visible global effect in the master R session when doing the parallel computing, this is (partially) true even for the `SerialParam`. 

With this principle in mind, if I'm a base-R user and have no idea how the seed is implemented in `BioParallel`, One day a friend asks me whether running `bplapply(x, function(x) runif(1))` can change the seed in the master,  my first intuition is definitely "no". As the master and workers are independent, there is no reason to assume the seed in the master can be changed. If he asks me whether running `bplapply(x, function(x) runif(1))` twice can get the same result, the answer should be "no" neither since the seed is not set for each worker. Therefore, my intuition as a user brings me to the following conclusions

1. The apply function(e.g. `bpapply` and `bpiterate`) should not change the global seed stream when called.
2. There should be an internal seed stream for preventing from getting the same result when calling the same code twice

Then, my proposal is kind of a mixture of Martin and Aaron

1. Randomness:     Every BPPARAM should have an internal seed stream that can be iterated when calling any apply functions
2. Reproducibility:  If `RNGseed` is set, use the `RNGseed` to initiate the internal stream
3. Fallback:             If `RNGseed` is not set, use the global seed instead

One last tricky question is when to initiate the seed stream. This might look obvious but actually we have two possibilities:

1. Initialize/reset the stream in `bpstart` only
2. Initialize the stream in the constructor of BPPARAM

These two possibilities adopt two different points of view. The former believes that a cluster must be a physical cluster, so the seed stream will be set when starting the cluster. However, the latter treats the BPPARAM as a conceptual cluster, that is, the seed stream always lives inside BPPARAM, turning the physical cluster on and off would not affect the seed stream. These two might looks similar but they are very different in practice. For example
```
p <- SnowParam()
bplapply(1:3, function(x) runif(1), BPPARAM = p)
bplapply(1:3, function(x) runif(1), BPPARAM = p)
```
For the first strategy, two `bpapply`s give the same results. However, the results are different for the second. I believe both are acceptable to the developer(That's why no one argues it under this topic), but as a user, I will vote for the second one since it is more intuitive and looks more natural.

LTLA:
> ```
> The task sent from the master will be executed in the worker
> ```

I don't think you can assume that. Certainly that wouldn't be the case with `SerialParam()`. Or you could imagine creating a `*Param` where the master performs execution of one of the tasks while the workers operate asynchronously. Maybe `MulticoreParam()` already does that, I don't know. But in any case, that's not the key takeaway for me.

> One day a friend asks me whether running `bplapply(x, function(x) runif(1))` can change the seed in the master

I would say "I don't know what to expect, and neither should you." And then I would tell them to turn on job-specific seeds by setting `RNGseed = NA` or whatever flag value we end up using for this. 

If I had _carte blanche_ to do whatever changes I wanted, then the best approach would be to hit the global stream so that the qualitative effect of the `runif()` call is respected (not the exact effect, but that's less important). Unfortunately, this may change code in the wild, so the best that can be done right now is to preserve the default behavior with a warning to suggest people opt-in to job-specific seeds if a change in per-job streams is detected.

I think it's fine to shift the global stream when it's explicitly requested. I don't think there's a need to complicate matters with another internal stream, especially given that the internal stream won't respond to `set.seed()`.

In any case, the current behavior is breaking a whole stack of my packages, so we should make a decision soon. An opt-in approach with a warning seems like an appropriately conservative solution to me.

Jiefei-Wang:
> I don't think you can assume that. Certainly that wouldn't be the case with `SerialParam()`. Or you could imagine creating a `*Param` where the master performs execution of one of the tasks while the workers operate asynchronously. Maybe `MulticoreParam()` already does that, I don't know. But in any case, that's not the key takeaway for me.

Obviously `SerialParam` would not fit this concept, that is why we must try our best to make `SerialParam` like a remote param. If the choice of the backend does matter to our result, how can we assume our code is reproducible when we do not have control over which backend our target audience will use? Just think about the example you proposed

> for example, a user might be expecting this to be reproducible:
> > library(BiocParallel)
> > set.seed(123); bplapply(1:5, function(x) { set.seed(x); runif(1) }); runif(1) # 0.2875775

If `SerialParam` can change the seed in master, then the result will depend on the backend. For `SerialParam`, the last `runif(1)` is equivalent to `set.seed(5); runif(2)[2]` . However, if you switch to `SnowParam`, it is equal to `set.seed(123); runif(1)`(if we assume the backend has already been started). The beauty of `BiocParallel` is that we can code once and run everywhere, now we will undermine it if the choice of backend kicks in our result.


> > One day a friend asks me whether running `bplapply(x, function(x) runif(1))` can change the seed in the master
>
> I would say "I don't know what to expect, and neither should you." And then I would tell them to turn on job-specific seeds by setting `RNGseed = NA` or whatever flag value we end up using for this.

Technically "I do not know" should be the correct answer, but why do we want to make it counterintuitive for the base-R user(e.g. users who are only familiar with `parallel`)? We do break some assumptions in the package development, but it only happens when the break can bring performance or coding benefits, which I fail to see in this case.

Anyway, no matter which direction we want to pursue, if we want to make it very different from the `parallel` package, I agree with Aaron that we should provide a fallback `RNGseed = NA` to make it less counterintuitive.

LTLA:
I don't know what you mean by "counterintuitive", because I don't even know what the intuitive approach would be here.

If you're going to fiddle with the seed, you're going to break someone's expectations of what's going to happen, one way or the other. You might be thinking in terms of abstractions with an arbitrary `BPPARAM`, but most users - and possibly even developers - are writing their code with a concrete `BPPARAM`. So changing the default seed behavior is going to change the behavior of existing code, in a way that is very difficult for the average user to diagnose (especially if it's buried deep in the stack).

Hence my strong preference for opt-in to any seed manipulations.

Jiefei-Wang:
> If you're going to fiddle with the seed, you're going to break someone's expectations of what's going to happen, one way or the other. You might be thinking in terms of abstractions with an arbitrary `BPPARAM`, but most users - and possibly even developers - are writing their code with a concrete `BPPARAM`. So changing the default seed behavior is going to change the behavior of existing code, in a way that is very difficult for the average user to diagnose (especially if it's buried deep in the stack).

No offense, but it looks like you misunderstand my argument. The default seed behavior(Bioc 3.13) is not to change the global seed, and that is what I'm talking about. You justify it from package maintenance, but I look this question from the user perspective: anything you can expect from the `parallel` package can (mostly) be expected from `BiocParallel`. I call it the "intuition" since I assume most people know `parallel` before `BiocParallel`, if there is any "surprise", we must have a good reason.

LTLA:
> The old seed behavior(Bioc 3.13) is not to change the global seed, and that is what I'm talking about.

Well, that's not quite true. The current seed behavior is simply undefined when `FUN` contains code that hits the global stream. Maybe it changes the global seed (`BPPARAM = SerialParam()`). Maybe it doesn't (`BPPARAM = SnowParam()`). Who knows?

> You justify it from package maintenance, but I look this question from the user perspective: anything you can expect from the `parallel` package can (mostly) be expected from `BiocParallel`. I call it the "intuition" since I assume most people know `parallel` before `BiocParallel`, if there is any ~surprise~ inconsistance, we must have a good reason.

I don't expect any similarity between **parallel** and **BiocParallel**. And neither should anyone else. If people want results that are exactly the same as those from **parallel**, then they should just use **parallel**.

Anyway, this discussion of consistency is irrelevant. Right now we're running out of time for the upcoming release. I suggest reverting all commits related to this issue so that (i) my packages build again and (ii) we don't break user-code in the wild.

mtmorgan:
The consistency of reproducibility across *Param / workers / tasks is a good feature. Introducing it is a breaking change whether introduced now or a month from now. So we're not going to undo all commits related to this issue.

https://github.com/Bioconductor/BiocParallel/pull/153 is a pull request that ignores the global random number stream for workers --

```
set.seed(123); result1 <- bplapply(1:3, runif)
set.seed(123); result2 <- bplapply(1:3, runif)
!identical(result1, result2)  # TRUE, just like bioc-3.13
```

and doesn't increment the global random number stream

```
set.seed(123); value <- runif(1)
set.seed(123); res <- bplapply(1:3, runif)
identical(value, runif(1)) # consistent across *Param, so better than bioc-3.13
```

Maybe a subsequent commit will introduce `RNGseed = NA` to mean 'use & increment (by 1, or fancier) the global RNG stream when the random number generator is used in a worker'.

mtmorgan:
I'll close this issue, with the solution as implemented in BiocParallel version 1.29.0 and documented for instance in the vignette https://bioconductor.org/packages/3.15/bioc/vignettes/BiocParallel/inst/doc/Random_Numbers.pdf. Open new issues with bug reports, etc.

PR #149, opened 2021-09-17
add force.GC / bpforceGC()

mtmorgan:
- set whether to force R garbage collection (expensive!) on every call to
  FUN()
- change default behavior -- force only for MulticoreParam
- improves https://github.com/Bioconductor/BiocParallel/pull/124

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: BiocParallel
 Type: Package
 Title: Bioconductor facilities for parallel evaluation
-Version: 1.27.8
+Version: 1.27.9
 Authors@R: c(
     person("Bioconductor Package Maintainer",
         email="maintainer@bioconductor.org", role="cre"),
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -68,6 +68,7 @@ export(
     bpstopOnError, "bpstopOnError<-",
     bpprogressbar, "bpprogressbar<-", bpRNGseed, "bpRNGseed<-",
     bptimeout, "bptimeout<-", bpexportglobals, "bpexportglobals<-",
+    bpforceGC, "bpforceGC<-",
     bplog, "bplog<-", bplogdir, "bplogdir<-",
     bpthreshold, "bpthreshold<-", bpresultdir, "bpresultdir<-",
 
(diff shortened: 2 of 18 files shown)


Jiefei-Wang:
Thanks for Solving the GC issue, I see 

https://github.com/Bioconductor/BiocParallel/blob/7e184a6bb0866a3086b546114f2b67e1bf4c17b5/R/MulticoreParam-class.R#L54

Is there any reason on why we have an `NA` value in `force.GC`? I think the default value in each constructor is good enough. If we just want to do the error check, isn't `length(force.GC) != 1L || !is.logical(force.GC)` better?

BTW, I do not expect we have a new exported method for turning the GC on and off since our user interface is already complicated and it might not be cost-effective to have an option for this simple function, but if we decide to pursue this direction, I propose we should make it more flexible: replace the switch with a counter, say `GC.counter`, the use case is

1. If `GC.counter = Inf`, it is equivalent to `force.GC = FALSE`
2. If `GC.counter = 1`, it is equivalent to `force.GC = TRUE`
3. If `GC.counter` is any integer greater than 1, do the GC after run `FUN`  `GC.counter` times.

Please let me your thoughts on it.

mtmorgan:
I simplified the interface, and use an existing function `.isTRUEorFALSE()` for validity.

Yes the `GC.counter` idea seems like a possibility but maybe difficult to implement (e.g., presumably this is relevant to _workers_, but we only have easy access to individual iterations of `FUN()` or perhaps _tasks_). Also, no one has asked for this...

Jiefei-Wang:
We can add a simple counter before 

https://github.com/Bioconductor/BiocParallel/blob/41aa8e8e146d86ade1a931607e3ed5f4b4d394af/R/ErrorHandling.R#L72

since this function has an enclosing environment and therefore we can keep track of how many times `FUN` is evaluated. Of course, this is not perfect since this counter only works within a task(e.g. a task can contain 3 `FUN` evaluations but `GC.counter` might be 4 and never be triggered), but at least this solution is simple enough and should work in most cases. I guess no one asks this as they even do not know our package will do GC in every `FUN`...

Other than that, I think the implementation is good enough, please feel free to update or merge with the master.

PR #150, opened 2021-09-20
Let bplapply and bpiterate share the same core function bploop.iterate

Jiefei-Wang:
This update lets `bplapply` and `bpiterate` share the same core function `bploop.iterate` and provides a new function `bpinit` to do some common works before `bploop`.  `SerialParam` will share the same `bploop` implementation with `SnowParam`. Here is a list of visible changes

1. Worker's seed is different from the current devel branch. In other words, this update does not promise reproducibility across different versions, but since we already break the reproducibility in the devel branch, I guess it is fine to break it twice.
2. No fallback to the SerialParam when `length(X) == 1`.
3. `bploop.lapply` and `bploop.iterator` are not backward compatible(They are exported functions), I did a search in Bioconductor code base and no other package uses them, so it should not break the existing packages.

Best,
Jiefei


R/bploop.R
```diff
@@ -252,14 +224,21 @@ bploop.lapply <-
...
+## - ARGS: the arguments to FUN
 bploop.iterate <-
-    function(manager, ITER, FUN, ARGFUN, BPPARAM, REDUCE, init, reduce.in.order,
-             ...)
+    function(manager, ITER, FUN, ARGS, BPPARAM, REDUCE, init, reduce.in.order)
 {
     cl <- bpbackend(BPPARAM)
 
+    seed <- .bpnextRNGstream(BPPARAM)
+    # seed <- .rng_next_substream(seed)
```
mtmorgan, line 240: The intention was that `.bpnextRNGstream()` was used on `bpstart()`, and the substreams were used for each element of `X` as mentioned in this commit bf9de5e1f01845c97d08c5b1fde8e31ce9b41eb3.

R/rng.R
```diff
@@ -1,3 +1,15 @@
+## Force bploop to iterate the seed stream n times
+.rng_bploop_iter <- function(n) {
+    structure(n, class = c(".rng_bploop_iter"))
+}
+
+## iterate the seed stream n times
+.rng_iterate_stream <- function(seed, n){
```
mtmorgan, line 7: rename `.rng_iterate_substream`

R/utilities.R
```diff
@@ -65,15 +65,68 @@
...
-    idx <- .splitIndices(length(X), tasks)
-    relist(X, idx)
+    ## If redo index presents, split X based on the index while
+    ## preserving the correct rng stream.
+    ## it will respect `tasks` value
+    if (length(redo_index)){
+        ## Two-pass algorithm, the first pass calculate
+        ## the required list size to allocate, then do
+        ## the real allocation
+        works_per_task <- ceiling(length(X)/max(tasks, 1L))
```
mtmorgan, line 122: Are these `elements_per_task`?

R/utilities.R
```diff
@@ -180,3 +233,16 @@
     )
     stop(msg, call. = FALSE)
 }
+
+.dummy_iter <- function(X){
```
mtmorgan, line 237: This seems like it's used in only one location, and could be moved and renamed `.bploop_lapply_iter`

R/ErrorHandling.R
```diff
@@ -134,6 +134,7 @@ bptry <- function(expr, ..., bplist_error, bperror)
 .error_bplist <- function(result) {
     idx <- which(!bpok(result) &
                   !vapply(result, inherits, logical(1), "not_available_error"))
+        # !vapply(result, inherits, logical(1), "unevaluated_error")
```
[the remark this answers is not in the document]
Jiefei-Wang: I originally want to ignore the `unevaluated_error` in `.error_bplist` as they are due to the error that occurs in the previous step and are confusing, but I think it might be better to do it in another pull request so I comment it out. I'll remove this line in the next commit.

R/bplapply-methods.R
```diff
@@ -45,58 +45,39 @@ setMethod("bplapply", c("ANY", "list"),
...
-        param <- as(BPPARAM, "SerialParam")
-        return(bplapply(X, FUN, ..., BPREDO=BPREDO, BPPARAM=param))
-    }
-
     ## which need to be redone?
     redo_index <- .redo_index(X, BPREDO)
     if (any(redo_index)) {
-        X <- X[redo_index]
-        compute_element <- redo_index
+        compute_element <- which(redo_index)
```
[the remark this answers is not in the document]
Jiefei-Wang: it is kind of stale as only the first version of `bplapply` requires it, but I think it could provide some performance benefit(but might be trivial) since subset by index should be faster than subset by the logical vector, so I just leave it here.

R/bplapply-methods.R
```diff
@@ -45,58 +45,39 @@ setMethod("bplapply", c("ANY", "list"),
...
-    cls <- structure(list(), class="lapply")
-    res <- bploop(cls, X, .rng_lapply, ARGFUN, BPPARAM)
+    if (length(compute_element)) {
+        BPREDO[compute_element] <- res
+        res <- BPREDO
+    }
 
     if (!is.null(res)) {
-        res <- do.call(unlist, list(res, recursive=FALSE))
         names(res) <- nms
```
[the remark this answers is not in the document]
Jiefei-Wang: Thanks for the catch! It will be fixed in the next commit.

R/bploop.R
```diff
@@ -177,73 +183,39 @@
...
-                running[d$node] <- .send_to(cl, d$node, value)
-            }
-        }
-    }
-
-    ## return results
-    if (!is.na(bpresultdir(BPPARAM)))
-        NULL
-    else result
+        ITER <- .dummy_iter(X)
```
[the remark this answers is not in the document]
Jiefei-Wang: Thanks! It will be reindented in the new commit.

R/bplapply-methods.R
```diff
@@ -45,58 +45,39 @@ setMethod("bplapply", c("ANY", "list"),
...
 
-    cls <- structure(list(), class="lapply")
-    res <- bploop(cls, X, .rng_lapply, ARGFUN, BPPARAM)
+    if (length(compute_element)) {
+        BPREDO[compute_element] <- res
+        res <- BPREDO
+    }
 
     if (!is.null(res)) {
-        res <- do.call(unlist, list(res, recursive=FALSE))
```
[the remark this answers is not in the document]
Jiefei-Wang: In the old implementation, `res` is a list of lists, so we need this line to unlist one layer of the list. For example
```
> res <- list(list(1,2), list(3,4))
> dput(do.call(unlist, list(res, recursive=FALSE)))
list(1, 2, 3, 4)
```
In the new implementation, `bplapply` depends on `bploop.iterate` and the iterator just return the result in a list format, so there is no need to unlist them again.

R/bpiterate-methods.R
```diff
@@ -24,34 +25,37 @@ setMethod("bpiterate", c("ANY", "ANY", "missing"),
...
-        on.exit(bpstop(BPPARAM), TRUE)
-    }
-
-    ## FUN
-    FUN <- .composeTry(
-        FUN, bplog(BPPARAM), bpstopOnError(BPPARAM),
-        timeout=bptimeout(BPPARAM), exportglobals=bpexportglobals(BPPARAM),
-        force.GC = bpforceGC(BPPARAM)
+    if(missing(init))
+        init <- substitute()
```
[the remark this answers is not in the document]
Jiefei-Wang: This should be stale code, it is used to explicitly construct the missing argument, but I think we do not need them now.

R/rng.R
Jiefei-Wang: Thanks, I will do it in the update

R/utilities.R
Jiefei-Wang: Yes, I call each element in the task a "work", but we can use "element" as well. I will rename it.

R/bploop.R
```diff
@@ -177,73 +183,39 @@
         else {
             lst <- as.list(env)
             idx <- setdiff(names(lst), c("reduce", "init", "value", "index"))
-            unname(lst[idx[order(as.integer(idx))]])
+            if(length(idx)){
+                as.list(do.call(c, unname(lst[idx[order(as.integer(idx))]])))
```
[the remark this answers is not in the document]
Jiefei-Wang: This is similar to `do.call(unlist, list(res, recursive=FALSE))` in your previous comment, but I think we can use `unlist` as well, I'll change it in the update.

R/utilities.R
Jiefei-Wang: Thanks, I'll rename it.

Jiefei-Wang:
Hi @mtmorgan ,

Thanks for the review and comments! That's really helpful. I have fixed most questions in the update. My only concern is your comment about the rng seed.

> The intention was that .bpnextRNGstream() was used on bpstart(), and the substreams were used for each element of X as mentioned in this commit bf9de5e.

If I understand the current devel implementation correctly, `.bpnextRNGstream` is not only used in `bpstart`. For example, in `bploop.iterate`, we have

https://github.com/Bioconductor/BiocParallel/blob/70b4d051a3f99db9c52a4bf70e88ed7cc631df5a/R/bploop.R#L271

`.rng_seeds_by_task` will call `.bpnextRNGstream` to obtain its initial seed. Since mine is inherited from the devel branch, it has a call to `.bpnextRNGstream` in the first place. 

BTW, I'm not sure it is unnecessary to iterate the rng stream again after getting the stream from `.bpnextRNGstream`, so I comment out L240 `seed <- .rng_next_substream(seed)`. That's why my branch cannot produce the same result as the devel branch using the same seed. Please let me know your thought about it.

Best,
Jiefei

R/bploop.R
```diff
@@ -268,20 +261,26 @@ bploop.iterate <-
...
+            list(X=X), list(FUN=FUN), ARGS,
+            list(BPRNGSEED = seed)
+        )
     ## initial load
     for (i in seq_len(workers)) {
         value <- ITER()
-        if (is.null(value)) {
+        if (inherits(value, ".rng_bploop_iter")) {
+            seed <- .rng_iterate_substream(seed, value)
+            value <- ITER()
```
mtmorgan, line 274: It looks like `ITER()` is called twice in the for loop?

R/bploop.R
mtmorgan: I think the use of `.bpnextRNGstream()` in the original code was an error, it should have been just `.RNGseed()`.

R/bploop.R
Jiefei-Wang: Yes, we need to make the iterator works with BPREDO, so I introduce a new class `.rng_bploop_iter` witch is merely designed to inform `bploop` to iterate the rng substream a few times. After the iteration, we should retrieve the next value from the iterator, which brings L274. There should be no two consecutive `.rng_bploop_iter` in the iterator, so we do not have to check the value class again.

R/bploop.R
Jiefei-Wang: Thanks, it looks like you did it on purpose, we need to memorize the seed used by the previous apply function. Otherwise two same apply calls will give the same result. There is a unit test for it

https://github.com/Bioconductor/BiocParallel/blob/70b4d051a3f99db9c52a4bf70e88ed7cc631df5a/inst/unitTests/test_rng.R#L196-L199

Since `.bpnextRNGstream` will iterate the seed and store it in the BPPARAM, it is able to pass the test, but `.RNGseed` cannot change the seed in BPPARAM and therefore cannot pass the test. 

But I do see a bug in my code caused by the fallback SerialParam in `bpinit`, I'll push a commit to fix it.

R/bploop.R
mtmorgan: Can you add a clarifying comment to the code?

Jiefei-Wang: No problem, I have updated the comments.

R/bploop.R
Jiefei-Wang: `.bpnextRNGstream` has been removed from the commit 50651fd as a side effect of the bug fixing anyway. `bpinit` is now able to memorize the seed stream used by the fallback SerialParam and set it back to the original *Param after `bploop`, this promises the result is consistent for the *Param with different worker numbers. For example
```
## This will fallback to the SerialParam, but it is still able to iterate its seed stream
> p <- SnowParam(workers = 1, RNGseed = 1)
> bpstart(p)
> as.numeric(bplapply(1, test, BPPARAM = p))
[1] 0.7720786
> as.numeric(bplapply(1, test, BPPARAM = p))
[1] 0.4301437
``` 
I do think there is still more work for the RNG seed stream, but at least with this new version we do not have to duplicate our implementation in many functions.

mtmorgan:
Can you please also format and comment the `.split_X_redo` code? Please follow the general style of the code, e.g., 
```
.split_X_redo <-
    function(X, redo_index, elements_per_task, n)
{
    ...
    for (i in seq_along(redo_index)) {
        redo <- redo_index[[i]]
        ...
        is_switch <- ifelse(
            is_switch || !redo, is_switch, x_len >= elements_per_task
        )
        if (!prealloc && is_switch) {
            if (last_redo_status ){
                splittedX[[task_i]] <- X[seq.int(x_start, length.out = x_len)]
            } else {
                splittedX[[task_i]] <- .rng_bploop_iter(x_len)
            }
        }
```

Jiefei-Wang:
Thanks, I have formatted and added comments to `.split_X_redo`. A unit test is added for the function as well.

mtmorgan:
Thanks @Jiefei-Wang for this. I'll merge the changes but would like the commit history to be tidied up into two, three, or four (?) commits that place the changes into logical groups, so that the history does not include changes that did not end up being included, and so that the logical steps required for the PR are apparent at some future point in time.

I'm sure there's a fancier way of doing what I want, but you might start a new branch from master (I just pushed a commit...), adding an 'export' of your current files (from https://stackoverflow.com/questions/2364147/how-to-get-just-one-file-from-another-branch/2364223 I think you can, on your new branch, `git restore --source=Jiefei-Wang-bploop .`), and then selectively commit with `git add --interactive`.

Please also take a moment, before starting the commits to the new branch, to review all code for the style changes I indicated in `.split_X_redo`, and also review the code base for code chunks (e.g., `.rng_seeds_by_task`) that are no longer used.

Jiefei-Wang:
Thanks, I'll do the final review and make another pull request.

PR #152, opened 2021-09-23
Let bplapply and bpiterate share the same core function bploop.iterate

Jiefei-Wang:
Hi @mtmorgan 

Here is a squashed version of the updates. The unused code is removed, all `if` and `for` statements should have the correct format now.

Best,
Jiefei


Jiefei-Wang:
wait a second, there are some stale unit tests.

Jiefei-Wang:
OK, I have removed the obsolete unit test. Since the last commit is also about updating the unit test so I just use force push to combine the changes. Not quite elegant, but works. Everything should look good now, please let me know if there is anything missing.

PR #154, opened 2021-10-01
Subsetting X using `[` instead of `[[`

Jiefei-Wang:
This pull request tries to avoid the use of `[[` on the variable `X` in `bplapply` and solves the following issue
```
.A = setClass("A", slots = c(a = "integer"))
setMethod("[", "A", function(x, i, j, ...) initialize(x, a = x@a[i]))
setMethod("length", "A", function(x) length(x@a))
as.list.A <- function(x, ...)
    lapply(seq_along(x), function(i) x[i])

x = .A(a = 1:3)
lapply(x, function(elt) elt@a)    # works because `lapply()` calls 'as.list' -- expensive
bplapply(x, function(elt) elt@a)  # fails because there is no `[[` method
```
Note that this can still fail for `SnowParam` as the worker does not have the definition of the class `A`(but works if the definition of `A` is in a package), so I edit `rng_apply` to make it fails gracefully. Another visible change is that I add `BPREDOIDX` to allow `.bploop_lapply_iter` to create task value on demand. This is more simple and efficient than using `.split_X_redo` to split tasks.


mtmorgan:
I'm concerned that at this stage in the release cycle we should be fixing known bugs rather than introducing new ones. Are all changes in the patch strictly necessary to address the regression?

Jiefei-Wang:
Thanks. I can cherry-pick the core changes from this pull request, and leave the rest to the next devel version. I'll close this now and open a new pull request

PR #155, opened 2021-10-02
Subsetting X using `[` instead of `[[`

Jiefei-Wang:
This pull request tries to avoid the use of `[[` on the variable `X` in `bplapply` and solves the following issue
```
.A = setClass("A", slots = c(a = "integer"))
setMethod("[", "A", function(x, i, j, ...) initialize(x, a = x@a[i]))
setMethod("length", "A", function(x) length(x@a))
as.list.A <- function(x, ...)
    lapply(seq_along(x), function(i) x[i])

x = .A(a = 1:3)
lapply(x, function(elt) elt@a)    # works because `lapply()` calls 'as.list' -- expensive
bplapply(x, function(elt) elt@a)  # fails because there is no `[[` method
```
Note that this can still fail for `SnowParam` as the worker does not have the definition of the class `A`(but works if the definition of `A` is in a package), so I edit `rng_apply` to make it fails gracefully.

### DIFF
--- a/R/bploop.R
+++ b/R/bploop.R
@@ -204,7 +204,7 @@
             i <<- i + 1L
             X[[i]]
         } else {
-            NULL
+            list(NULL)
         }
     }
 }
@@ -274,7 +274,8 @@ bploop.iterate <-
 
     ARGFUN <- function(X, seed)
         c(
-            list(X=X), list(FUN=FUN), ARGS,
+            list(X=X, BPELEMENTS = length(X)),
+            list(FUN=FUN), ARGS,
             list(BPRNGSEED = seed)
         )
     ## initial load
@@ -288,7 +289,7 @@ bploop.iterate <-
             seed <- .rng_iterate_substream(seed, value)
             value <- ITER()
         }
-        if (is.null(value[[1]])) {
+        if (identical(value, list(NULL))) {
             if (i == 1L)
                 warning("first invocation of 'ITER()' returned NULL")
             break
@@ -329,7 +330,7 @@ bploop.iterate <-
             seed <- .rng_iterate_substream(seed, value)
             value <- ITER()
         }
-        if (!is.null(value[[1]])) {
+        if (!identical(value, list(NULL))) {
             i <- i + 1L
             value_ <- .EXEC(i, .rng_lapply, ARGFUN(value, seed))
             running[d$node] <- .send_to(cl, d$node, value_)
--- a/R/rng.R
+++ b/R/rng.R
@@ -133,11 +133,23 @@
 ## lapply, but with 'FUN()' wrapped so that each call uses a new
 ## random number stream
 .rng_lapply <-
-    function(X, FUN, ..., BPRNGSEED)
+    function(X, FUN, ..., BPRNGSEED, BPELEMENTS)
 {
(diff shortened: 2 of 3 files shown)


Jiefei-Wang:
Hi Martin, thanks for the merge. I have to reopen this pull request since there is a problem in your commit 20fc1159c6fe99d9fc537ef7843a1a4e42d0b76b, the number of elements in `bplist_error` does not match the number of elements in `X`. For example
```
.B <- setClass("B", slots = c(b = "integer"))
setMethod("[", "B", function(x, i, j, ...) initialize(x, b = x@b[i]))
setMethod("length", "B", function(x) length(x@b))
as.list.B <<- function(x, ...) lapply(seq_along(x), function(i) x[i])
x <- .B(b = 1:3)

res <- tryCatch(
        bplapply(x, function(elt) elt@b, BPPARAM=SnowParam(workers = 2)),
        error = identity
)


## [1] FALSE
```

I think the core issue is that we do not know the number of elements in a task when there is an error in `lapply` in `.rng_lapply`. That's why I have to pass the length of the task along with the other task values. Otherwise, we will have a mismatch between `X` and `BPREDO`. I'll propose we revert the changes in `.rng_lapply` in 20fc1159c6fe99d9fc537ef7843a1a4e42d0b76b and add the following line at L125 in `test_bplapply.R`
```
checkIdentical(length(x), length(attr(res, "result"))
```


Best,
Jiefei

mtmorgan:
I saw the mismatch and added help page documentation, in the commit, indicating that the length of the return value did not match the length of the input. I did this because (a) we are close to release and I want to be very conservative about changes we introduce (b) reporting the error actually represents a bug fix compared to the release (the release branch fails with a crytpic error) rather than a regression related to use of `[[` (c) other '*_comm' errors also appear likely to be broken, and we have not addressed these systematically and (d) the current patch is not correct
```
f <- function(n, workers, tasks) {
    .A = setClass("A", slots = c(a = "integer"))
    setMethod("[", "A", function(x, i, j, ...) initialize(x, a = x@a[i]))
    setMethod("length", "A", function(x) length(x@a))
    as.list.A <<- function(x, ...)
        lapply(seq_along(x), function(i) x[i])

    x = .A(a = seq_len(n))
    lapply(x, function(elt) elt@a)
    bplapply(x, function(elt) elt@a, BPPARAM = SnowParam(workers, tasks = tasks))
}
> length(bptry(f(11, 2, 3)))
## [1] 7
```
Also, I didn't like the code duplication (`sapply(sys.calls()...`) which suggests that the solution isn't integrated into the overall error handling infrastructure (maybe there should be a new or updated error class added to ErrorHandling.R?) and didn't understand the first branch of the `if()`, which doesn't report an error even if an error occurred (but how could it occur, since the length of X is 0? I would have skipped the conditional and written `c(list(.error_remote(e, call)), rep(list(.error_unevaluated()), BPELEMENTS - 1L))` which covers everything?).

Jiefei-Wang:
Thanks for the explanation! If this is documented then it should be fine. I understand the conservativeness given that we are approaching the deadline. Maybe we can come up with a comprehensive solution for this issue in the next devel? I'll close this for now and give it more thought.

PR #156, opened 2021-10-03
Transient multicore param regression

mtmorgan:
This functionality was available prior to recent changes. It's important because it can provide significant speed-up when there is a lot of data, and for the common use case where the `BPPARAM` has not been started by the user.
```
f <-
    function(param, y)
{
    bplapply(1:100, function(x, y) sum(y), y = y, BPPARAM = param)
}

y <- rnorm(1000000)
param = MulticoreParam(tasks = 100)
system.time(f(param, y)) # less than 1s
param <- bpstart(MulticoreParam(tasks = 100))
system.time(f(param, y)) # more than 4s
```

### DIFF
--- a/R/TransientMulticoreParam-class.R
+++ b/R/TransientMulticoreParam-class.R
@@ -17,9 +17,7 @@ setMethod(
     "bpstart", "TransientMulticoreParam",
     function(x, ...)
 {
-    .rng_internal_stream$set()
     parallel::mccollect(wait=TRUE)
-    .rng_internal_stream$reset()
 
     rm(
         list=ls(envir = .TRANSIENTMULTICOREPARAM_JOBNODE),
@@ -87,9 +85,7 @@ setMethod(
             for (id in names(result))
                 .BUFF[[id]] <- result[[id]]
         }
-        .rng_internal_stream$set()
-        id <- sample(names(.BUFF), 1L)
-        .rng_internal_stream$reset()
+        id <- head(names(.BUFF), 1L)
 
         value <- .BUFF[[id]]
         rm(list = id, envir = .BUFF)
--- a/R/bpinit.R
+++ b/R/bpinit.R
@@ -1,9 +1,11 @@
 .bpinit <-
     function(manager, FUN, BPPARAM, ...)
 {
+    ## coercions to avoid serializing to the workers
     fallback <- FALSE
     if (!inherits(BPPARAM, "SerialParam")) {
         if (!bpschedule(BPPARAM) || bpnworkers(BPPARAM) == 1L) {
+            ## use SerialParam when only one worker
             fallback <- TRUE
             oldParam <- BPPARAM
             BPPARAM <- as(BPPARAM, "SerialParam")
@@ -12,6 +14,13 @@
 
     ## start / stop cluster
     if (!bpisup(BPPARAM)) {
+        ## use TransientMulticoreParam when MulticoreParam has not
+        ## started
+        if (is(BPPARAM, "MulticoreParam")) {
+            fallback <- TRUE
+            oldParam <- BPPARAM
+            BPPARAM <- TransientMulticoreParam(BPPARAM)
+        }
         BPPARAM <- bpstart(BPPARAM)
         ## the fallback SerialParam must inherit the seed stream from
         ## BPPARAM


Jiefei-Wang:
Thanks for bringing this problem up. It's on my list but I haven't touched it yet. It looks like the implementation of the transient multicore param is good since it does not depend on the iterator of `X`, I would not have any comment for it. Please feel free to merge it.

Best,
Jiefei

Jiefei-Wang:
Hi Martin, I see you have made more changes to `bpinit`. These changes the behavior of the function. We previously fallback to SerialParam when the worker number is less than 1, no matter whether the cluster has been started or not. With the latest commit, we only use the fallback SerialParam when the user-provided BPPARAM is not started. This may bring some side effects. For example, if the worker number is 0 but BPPARAM has been started, we do not fall back to the SerialParam as we use to

https://github.com/Bioconductor/BiocParallel/blob/e2f9f01617faedda925f2839e7012d7c9d2c9e62/R/bpinit.R#L4-L10

This might be impossible for SnowParam as we have a fixed number of workers, but it can happen for the PARAM with a dynamic worker number(RedisParam). `bpschedule` can affect the fallback mechanism in the same way as `bpnworkers`. I suggest we should revert the change in `bpinit` to avoid these side effects.

Best,
Jiefei

mtmorgan:
I think it was wrong to fall back to SerialParam whenever the worker number was 1; instead we should only fall back if the cluster has not been started (e.g., wanting to start and load packages in a single worker, independently of the main process, and use that independent environment on subsequent calls to bplapply() etc.). Allow for `bpnworkers(BPPARAM) == 0` on line 4?

I didn't mean to change the logic of
```
-    if (!inherits(BPPARAM, "SerialParam")) {
-        if (!bpschedule(BPPARAM) || bpnworkers(BPPARAM) == 1L) {
```
when I wrote
```
+        if (!inherits(BPPARAM, "SerialParam") &&
+            (!bpschedule(BPPARAM) || bpnworkers(BPPARAM) == 1L)) {
```
Did I mess things up?

Jiefei-Wang:
> I think it was wrong to fall back to SerialParam whenever the worker number was 1; instead we should only fall back if the cluster has not been started

I agree that we should not fall back to SerialParam when the worker number is 1. I just think we need the fallback when `bpnworkers(BPPARAM) == 0L` or `bpschedule(BPPARAM) == FALSE` even when `bpisup(BPPARAM) == TRUE`. 

> I didn't mean to change the logic of ...

There is nothing wrong at L9-10. The problem is that if `bpisup(BPPARAM) == TRUE`, L4 will bypass the fallback condition at L9-10...

mtmorgan:
Does this update address the problem (I find the structure much easier to follow, where each 'special condition' is handled in a logical unit, rather than split between multiple conditionals and the already-complicated-to-reason-about on.exit handlers)? 

```
 .bpinit <-
     function(manager, FUN, BPPARAM, ...)
 {
-    if (!bpisup(BPPARAM)) {
-        ## start cluster.
-
-        ## These conditions avoid cost of serializing data to workers
-        ## when not necessary
+    ## Conditions for starting a cluster, or falling back to (and
+    ## starting) a SerialParam
+    start_or_fall_back <-
+        !bpisup(BPPARAM) ||          # cluster hasn't started
+        bpnworkers(BPPARAM) == 0L || # e.g., in dynamic cluster like RedisParam
+        !bpschedule(BPPARAM)         # e.g., in nested parallel call
+    if (start_or_fall_back) {
         if (!inherits(BPPARAM, "SerialParam") &&
-            (!bpschedule(BPPARAM) || bpnworkers(BPPARAM) == 1L)) {
+            (!bpschedule(BPPARAM) || bpnworkers(BPPARAM) <= 1L)) {
+            ## use SerialParam when zero workers (e.g., in a dynaamic
+            ## cluster like RedisParam, where there are no registered
+            ## workers) or one worker (when the cost of serialization
+            ## to a remote worker can be avoided), or when
+            ## bpschedule() is FALSE (e.g,. because we are already in
+            ## a parallel job).
```
(it looks like `bpschedule()` is broken [e.g., the example on `?bpschedule` does not match the comment]) but that can be addressed separately; it is not a regression).

Jiefei-Wang:
Thanks. A minor comment is that we can cache and reuse the value of `bpnworkers(BPPARAM)` as it might involve a network request. Other than that the rest should be good enough.

mtmorgan:
My next attempt is
```
    nworkers <- bpnworkers(BPPARAM) # cache in case this requires a netowrk call
    fallback_condition <-
        !inherits(BPPARAM, "SerialParam") &&
        nworkers == 0L ||    # e.g., in dynamic cluster like RedisParam
        !bpschedule(BPPARAM) # e.g., in nested parallel call
    if (!bpisup(BPPARAM) || fallback_condition) {
        if (fallback_condition || nworkers == 1L) {
```

Jiefei-Wang:
Thanks, I think this looks much better. I do not have more comments on it. Please feel free to push and close this topic.

PR #158, opened 2021-10-05
make BPREDO reproducible

Jiefei-Wang:
The current BPREDO does not handle the rng seed stream correctly and the result will not be reproducible. For example
```
FUN <- function(i) rnorm(1)
FUN0 <- function(i) {
    if (identical(i, 7L)) {
        stop("i == 7")
    } else rnorm(1)
}

param <- SerialParam(RNGseed = 123, stop.on.error = FALSE)
bpstart(param)
target <- unlist(bplapply(1:11, FUN, BPPARAM = param))
bpstop(param)

bpstart(param)
## Something is wrong here
result <- bptry(bplapply(1:11, FUN0, BPPARAM = param))
## redo the apply function
result_redo <- unlist(bplapply(1:11, FUN, BPREDO = result, BPPARAM = param))

## the result is not the same as the user expect
identical(result_redo, target)
```
This pull request fixes the issue by attaching the initial seed stream used by BPPARAM to the erroneous result and recover the stream when `BPREDO` is used. Also, if `BPREDO` is presented, the seed stream will not be iterated after the apply function.

### DIFF
--- a/R/bpinit.R
+++ b/R/bpinit.R
@@ -66,7 +66,10 @@
     }
 
     if (!all(bpok(res))) {
-        attr(res, "BPREDOSEED") <- BPREDOSEED
+        ## attach the seed only when no
+        ## BPREDO presents
+        if (!length(BPREDO))
+            attr(res, "BPREDOSEED") <- BPREDOSEED
     } else {
         attr(res, "BPREDOSEED") <- NULL
     }
--- a/R/bpinit.R
+++ b/R/bpinit.R
@@ -1,5 +1,5 @@
 .bpinit <-
-    function(manager, FUN, BPPARAM, ...)
+    function(manager, FUN, BPPARAM, BPREDO = list(), ...)
 {
     if (!bpisup(BPPARAM)) {
         ## start cluster.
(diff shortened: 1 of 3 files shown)


mtmorgan:
This looks good. It also makes BPREDO work when the seed is not specified

```
FUN <- function(i) sum(.Random.seed)
FUN0 <- function(i) {
    value <- sum(.Random.seed)
    if (identical(i, 7L))
        stop(value)
    value
}

param <- SerialParam(stop.on.error = FALSE)
result <- bptry(bplapply(1:11, FUN0, BPPARAM = param))
result_redo <- unlist(bplapply(1:11, FUN, BPREDO = result, BPPARAM = param))
identical(result_redo[[7]], as.numeric(conditionMessage(result[[7]])))
```
I'll wait for resolution of https://github.com/Bioconductor/BiocParallel/pull/156 before committing

Jiefei-Wang:
Thanks, I did not consider the no-seed case at all. It is nice we can make this case reproducible as well. I made a minor change to this branch to avoid reattaching the incorrect seed when BPREDO is given.

mtmorgan:
I didn't catch the fix before pushing. Is this an optimization, or a bug fix?

Jiefei-Wang:
It's a bug fix. We should not attach the seed from the BPPARAM when BPREDO is provided by the user. This can happen if the user-provided BPREDO leads to an error again. It is an edge case though.

PR #159, opened 2021-10-05
attach `BPREDOSEED` to the result only when the attribute does not exist

Jiefei-Wang:
Hi Martin, this is the one that is left in the previous pull request. We should not attach `BPREDOSEED` to the result when the result contains the attribute BPREDOSEED.  Otherwise, it can overwrite the existing `BPREDOSEED`.

Best,
Jiefei

### DIFF
--- a/R/bpinit.R
+++ b/R/bpinit.R
@@ -73,7 +73,8 @@
     }
 
     if (!all(bpok(res))) {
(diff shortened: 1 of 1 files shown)


Jiefei-Wang:
close with the commit aa6d054a8fe79d21c4569976799067c7447bd703

PR #164, opened 2021-10-22
[Bioc 3.15] Implementing the 6-layer model for the apply function

Jiefei-Wang:
This pull request follows the pull roadmap in https://github.com/Bioconductor/BiocParallel/pull/162 and implements the 6-layer model for the apply function. The changes include:

1. The function `.bploop_lapply_iter` is revised such that there is no need to split `X` in advance, the argument `X` for `FUN` will be generated on demand.  
2. `.redo_index` will return the index of `X` instead of a logical vector. 
3. A minor bug fix for the missing log and result issue when a worker error occurs.

Best,
Jiefei


Jiefei-Wang:
The task manager has been revised to use the environment to manage the workers.

PR #157, opened 2021-10-04
accept `bplist_error` as the argument of BPREDO

Jiefei-Wang:
Allow `bplist_error` object to be used directly as the argument of `BPREDO`, for example
```
f = sqrt
x = list(1, "2", 3)
x.fix = list(1, 2, 3)

res <- tryCatch(bplapply(x, f), error=identity)
bplapply(x.fix, f, BPREDO=res)
```

PR #167, opened 2021-11-08
implement bpresult

mtmorgan:
replaces https://github.com/Bioconductor/BiocParallel/pull/157


Jiefei-Wang:
Thanks for the revision. I'm fine with the function renaming, but I'm concerned that changing the class of the output can lead to an unexpected result since `bpresult` is merely an S3 wrapper, but we cannot assume the output is always an S3. For example
```
.A <- setClass("A", contains = "list")
setMethod("show", "A", function(object) message("HI"))

## expected output: .A()
## actual output: list()
bpiterate(function()NULL, identity,
                 REDUCE = function(x,y)x,
                 init = .A(),
                 BPPARAM = SerialParam())
```
I'll suggest do not add the wrapper to the output, at least for `bpiterate`.

mtmorgan:
Would this be addressed by providing the wrapper in the default REDUCE, rather than in `bpiterate()` per se? What about only making the return value a class when there is an error? I don't think it's a good idea to show the user the BPREDOSEED attribute.

Jiefei-Wang:
> Would this be addressed by providing the wrapper in the default REDUCE, rather than in `bpiterate()` per see? 

Could you please provide an example of it? 

> What about only making the return value a class when there is an error? I don't think it's a good idea to show the user the BPREDOSEED attribute.

Isn't it the purpose of `bperror_list`? By calling `bpresult` I think the user wants to see the intermediate result, not another class of object. There will be no `BPREDOSEED` in the future, we will use an environment `REDOENV` to hide everything needed for the `BPREDO`.

Another idea can be using `bperror_list` as the `BPREDO` argument. Calling `bpresult` on `bperror_list` can give the intermediate result, but it should not be passed to `BPREDO`. By doing this we can still redo the evaluation and give the user a chance to see the "pure" output, but this also means some unlucky developers need to revise their package code to adapt to the changes. However, by peeking the Bioconductor code base this seems doable, only a few packages would be affected.

https://code.bioconductor.org/search/search?q=BPREDO+-r%3ABiocParallel

PR #180, opened 2021-12-28
implement the task cache for the SnowParam

Jiefei-Wang:
I guess we already found a good reason why we need to cache the task. From our [test code](https://gist.github.com/mtmorgan/8886fba08bc634e4408cb73c65ac5a47). The `SnowParam` takes more than 300 seconds without the cache, but it only takes 17 seconds with the cache.

The only concern is that by caching the user-supplied `FUN`, we assume all objects in the enclosing environment of `FUN` must be constant. Otherwise, it may lead to an incorrect result. After thinking about this for days, I decide not to add an argument to control the cache behavior as the order that the task gets finished is not assured. It is possible to get two task results reduced before we sent out a new task. Therefore, dynamically changing the function data in the reducer can cause more trouble than it is worth. In our test, we just change the data in the iterator and it already makes things very hard to debug, so I consider changing any objects in the enclosing environment that `FUN` depends on is a poor design. Therefore, we do not need to support it.

### DIFF
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -127,5 +127,7 @@ export(
     .manager, .manager_send, .manager_recv,
     .manager_send_all, .manager_recv_all,
     .manager_capacity,
-    .manager_flush, .manager_cleanup
+    .manager_flush, .manager_cleanup,
+    .task_type, .EXEC_static,
+    .EXEC_dynamic, .remake_EXEC
 )
--- a/R/DeveloperInterface.R
+++ b/R/DeveloperInterface.R
@@ -1,9 +1,10 @@
 ##
 ## see NAMESPACE section for definitive exports
 ##
+## Manager class
+.TaskManager <- setClass("TaskManager", contains = "environment")
 
 ## server
-
 setGeneric(
     ".send_to",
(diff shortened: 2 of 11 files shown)


mtmorgan:
I haven't looked at this closely, but the use case is actually driven by RedisParam so I'm wondering if the solution is specific to SnowParam as suggested in the title, and if so whether it could be generalized?

Also, one formatting thing about your pull requests is the horizontal trailing white space, e.g., at the ends of the lines
```
.load_EXEC_static <- 
    function(EXEC) 
```
In RStudio I think there is an option to automatically 'remove trailing whitespace on save' or something similar.

Jiefei-Wang:
Sorry for the confusion, this pull request is not specific to `SnowParam`, it works for the other backends as well. We export four developer functions to allow the other backends to cache the task, `SnowParam` is just the first backend that has the task cache enabled.

As for the formatting issue, I have enabled the whitespace stripping and removed the space from the functions. I made a force update to keep the commit history simple.

Jiefei-Wang:
There are some minor bugs with the cache implementation, so please hold on when I'm testing `RedisParam`.

Jiefei-Wang:
temporary close it.

PR #181, opened 2021-12-30
fix a validation issue for BiocParallelParam

Jiefei-Wang:
I cannot believe no one sees this bug in practice. Perhaps that's because we do not really return the worker names from `bpworkers` in all common `PARAM`s, but `RedisParam` does, and I see it...

### DIFF
--- a/R/BiocParallelParam-class.R
+++ b/R/BiocParallelParam-class.R
@@ -88,8 +88,8 @@ setValidity("BiocParallelParam", function(object)
 
     if (is.character(workers)) {
         if (length(workers) < 1L)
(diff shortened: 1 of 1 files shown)


mtmorgan:
Does this lead to an actual error, rather than just an inefficient use of resources?

Jiefei-Wang:
It will cause an actual error when `bpworker` returns a character worker vector. `SnowParam(workers = rep("localhost", 3), tasks =2)` is a stable way to trigger it. In `RedisParam`, I found this error when I was trying to call a `bp* <-` function. The replacement function will call the validation method and therefore throw an error.

mtmorgan:
Closed by 273ddc7dc937d57655f27a130a4c692a12740408

PR #182, opened 2022-01-02
Add task cache feature and enable it by default for the SnowParam

Jiefei-Wang:
This pull requests contains three new developer APIs: `.task_static`, `.task_dynamic`, and `.remake_task`. The use of these three functions is optional. All existing `Param`s work with this new version, they just cannot gain any benefit from the task cache feature.

### DIFF
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -127,5 +127,6 @@ export(
     .manager, .manager_send, .manager_recv,
     .manager_send_all, .manager_recv_all,
     .manager_capacity,
-    .manager_flush, .manager_cleanup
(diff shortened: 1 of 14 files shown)


mtmorgan:
I haven't looked at the pull request, but I think the general 'API' rule should group related tasks lexicographically, so `.task_remake`

Jiefei-Wang:
Thanks for your suggestion, I have made this change. I also change `.task_static` to `.task_const` as I think the former one might be too close to C language, so the new APIs are `.task_const`, `.task_dynamic`, and `.task_remake`.

PR #184, opened 2022-02-16
allow automatically exporting the global variables used in the function

Jiefei-Wang:
This pull request supports auto export variables/packages from the manager to the worker when `FUN` in `bplapply` requires it. 

Use case 1
```
x <- "hello"
bplapply(1, function(i) x, BPPARAM = SnowParam(2))
# [[1]]
# [1] "hello"
```

Use case 2
```
library(S4Vectors)
bplapply(1, function(i) DataFrame(), BPPARAM = SnowParam(2))
# [[1]]
# DataFrame with 0 rows and 0 columns
```


Jiefei-Wang:
I did a minor cleanup to remove the empty lines.

PR #187, opened 2022-02-19
Check the port status before creating the cluster

Jiefei-Wang:
When the user does not specify the port number, `SnowParam` will randomly choose the port without verifying whether the port is used or not, so it might select a port that is being used by the other process. This pull request alleviates the problem by adding a port check before returning the random port number.


mtmorgan:
I think this should be more informative to the user, e.g.,
```
Failed to connect on port xxx, trying again...
Failed to connect on port yyy, trying again...
```
and after say five failures fail hard with with `.stop("cannot find open port; see <relevant section in vignette or help pages>")`.

Jiefei-Wang:
Thanks, do you mean we should give a message/warning when the user specifies the port number but the number is occupied? I feel that when the user uses the default setting, this should be a part of the port searching procedure and we do not have to get the user's attention unless there is any error. Please let me know your thought.

mtmorgan:
In the patch I just saw an infinite loop? Which doesn't sound like a good idea... or did I misread the code?

mtmorgan:
I think the message should be printed each time the port selection fails.

Jiefei-Wang:
Thanks, I have changed it to a for loop and give a message when the port is not opening.

PR #190, opened 2022-03-02
[patch 1] Add 'bpoptions'

Jiefei-Wang:
After our discussion on Tuesday, I think it is better to set `fallback` as an attribute of `BPPARAM`, so this pull request only contains the vanilla `bpoptions`. `fallback` and `exports` will be added in the other pull requests.


mtmorgan:
merged

PR #201, opened 2022-04-20
Do not include the flag -lrt on windows

Jiefei-Wang:
'librt' is not used by 'BiocParallel' on Windows, so I removed it.


mtmorgan:
The build report https://bioconductor.org/checkResults/devel/bioc-LATEST/BiocParallel/palomino3-install.html  does not show use of `-lrt` on windows
```
g++ -shared -s -static-libgcc -o BiocParallel.dll tmp.def ipcmutex.o -lbcrypt -LC:/rtools42/x86_64-w64-mingw32.static.posix/lib/x64 -LC:/rtools42/x86_64-w64-mingw32.static.posix/lib -LF:/biocbuild/bbs-3.15-bioc/R/bin/x64 -lR
```
and since there is no `configure.win` the default Makevars
```
$ cat Makevars
PKG_LIBS =
```
is used. Maybe you have an artifact from using the same source tree for Windows Linux subsystem and Windows?

mtmorgan:
The build report https://bioconductor.org/checkResults/devel/bioc-LATEST/BiocParallel/palomino3-install.html  does not show use of `-lrt` on windows
```
g++ -shared -s -static-libgcc -o BiocParallel.dll tmp.def ipcmutex.o -lbcrypt -LC:/rtools42/x86_64-w64-mingw32.static.posix/lib/x64 -LC:/rtools42/x86_64-w64-mingw32.static.posix/lib -LF:/biocbuild/bbs-3.15-bioc/R/bin/x64 -lR
```
and since there is no `configure.win` the default Makevars
```
$ cat Makevars
PKG_LIBS =
```
is used. Maybe you have an artifact from using the same source tree for Windows Linux subsystem and Windows?

Jiefei-Wang:
I see, thanks for the explanation. The Win and the Linux subsystem do use the same source, so I thought we have a Makevars in BiocParallel... Close now

PR #203, opened 2022-04-21
Rewrite DoparParam using the manager interface

Jiefei-Wang:
The new implementation makes the `DoparParam`  behaves much similar to the other `BPPARAM`, like the way the job gets divided, the RNG seed used by each task. 

Because the manager interface is at the lowest level, there is no need to maintain two different versions of `bplapply`.


mtmorgan:
Merged with 450368a2492ae7eb9f40b6b5e373d8e0c64a78e4

Issue #234, opened 2022-11-22
exported entities are destroyed in caller


vjcitn:
```
> chkd = "a"
> chkd
[1] "a"
> bplapply(1:2, function(x) x, BPOPTIONS=bpoptions(exports=c("chkd")))
[[1]]
[1] 1

[[2]]
[1] 2

> chkd
Error: object 'chkd' not found
No suitable frames for recover()
```
feature or bug?

vjcitn:
Only with SerialParam()?

vjcitn:
I don't observe it with SnowParam or MulticoreParam

mtmorgan:
I vote bug, @Jiefei-Wang ?

Jiefei-Wang:
Yes, it is. I'll fix it.

PR #235, opened 2022-11-22
Fix issue 234

Jiefei-Wang:
Recover the replaced variables in the global environment after the worker finishes its job. Fix issue https://github.com/Bioconductor/BiocParallel/issues/234

### DIFF
--- a/R/worker.R
+++ b/R/worker.R
@@ -238,12 +238,20 @@
         suppressPackageStartupMessages(library(pkg, character.only = TRUE))
     }
     ## Add variables to the global space and remove them afterward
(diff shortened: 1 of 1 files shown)


DarwinAwardWinner:
As a minor optimization, you could probably skip the `on.exit` if `replaced_variables` is empty.

Jiefei-Wang:
Thanks for your feedback. The main purpose of the `on.exit` function is to do a cleanup at the end of the execution, so the `on.exit` function must be used no matter if `replaced_variables` is empty or not...

Issue #93, opened 2019-02-07
Need to link to -lrt for shm_open


spotrh:
In Fedora, I noticed that our builds of BiocParallel were failing, but only on ppc64le and aarch64 architectures. The failure looks like this:

```
Error: package or namespace load failed for 'BiocParallel' in dyn.load(file, DLLpath = DLLpath, ...):
 unable to load shared object '/builddir/build/BUILDROOT/R-BiocParallel-1.16.5-1.fc30.ppc64le/usr/lib64/R/library/BiocParallel/libs/BiocParallel.so':

/builddir/build/BUILDROOT/R-BiocParallel-1.16.5-1.fc30.ppc64le/usr/lib64/R/library/BiocParallel/libs/BiocParallel.so:
undefined symbol: shm_open
Error: loading failed
```

Did some digging, and R was not linked to librt (which has shm_open) because R only links to librt if this configure check passes:

`AC_CHECK_LIB(rt, clock_gettime)`

Sure enough, the Fedora 30 builds of glibc for ppc64le and aarch64 do not have clock_gettime in librt.so.

Since it is not safe to assume that R is linked to librt, you really need to add a configure script that checks for it and adds it to src/Makevars when needed. I'm going to submit a PR to add this.

spotrh:
The other way to resolve this is to get upstream R changed to always link to librt, but since R doesn't need librt except when nothing else provides clock_gettime (as far as I can see), I'm not sure if they'd be willing.

spotrh:
Thanks for the quick turnaround!

PR #236, opened 2022-11-26
update configure.ac to use C++

mtmorgan:
- test for sys/mman.h
- use AC_SEARCH_LIBS() to add -lrt if necessary, instead of AC_CHECK_LIB(), as recommended by https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Libraries.html
- updates #93, updates #94


mtmorgan:
@spotrh Not sure if you still have a configuration where you can test this update? It uses C++ as configured by R when performing the test. It also tests for the availability (in C++) of the appropriate header file. This is from a comment from a CRAN member.

mmokrejs:
I am hitting this with `R-3.4.4` on `Ubuntu-20.04` on which I am trying to install `Bioconductor-3.6`. I do not see `configure.ac` to be patched in the `BiocParallel_1.12.0.tar.gz` file.

But magic happened:

```
$ R CMD INSTALL BiocParallel_1.16.6.tar.gz 
* installing to library ‘/home/biomex/lib64/R/library’
* installing *source* package ‘BiocParallel’ ...
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking for shm_open in -lrt... yes
configure: creating ./config.status
config.status: creating src/Makevars
** libs
g++ -std=gnu++11 -I/home/biomex/lib64/R/include -DNDEBUG  -I"/home/biomex/lib64/R/library/BH/include" -I/usr/local/include   -fpic  -g -O2 -c ipcmutex.cpp -o ipcmutex.o
g++ -std=gnu++11 -shared -L/home/biomex/lib64/R/lib -L/usr/local/lib64 -o BiocParallel.so ipcmutex.o -lrt -L/home/biomex/lib64/R/lib -lR
installing to /home/biomex/lib64/R/library/BiocParallel/libs
** R
** inst
** preparing package for lazy loading
** help
*** installing help indices
** building package indices
** installing vignettes
** testing if installed package can be loaded
* DONE (BiocParallel)
$
```

So the `-lrt` got into its place.

mtmorgan:
@mmokrejs thanks for the report of success with BiocParallel_1.16.6.

[Bioconductor-3.6 is from 2017](https://bioconductor.org/about/release-announcements/) and is very old. Changes are introduced into the 'devel' or sometimes current release branch, but cannot be ported to earlier releases (Bioconductor does not allow commits to older releases, and does not build & propagate older packages).

Issue #241, opened 2023-02-24
Troubles with bplapply within function (using SnowParam on Windows)


DavoSam:
Hello,

I am new to BiocParallel and am trying to write a function that performs subsampling of a SingleCellExperiment object. The function below generates a list of randomly sampled indices based on the number of columns of the input object:

```R
dummyFUN2 = function(ob,n,p,n_cores=2) {
  BP = BiocParallel::bpparam()
  BP$workers = n_cores
  bplapply(1:n, 
           function(y, ob, p, r) { len = ncol(ob); sample(x = 1:len, size = p*len, replace = r) }, 
           ob = ob, p = p, r = FALSE, BPPARAM = BP)
  
}

```
I tried to be careful about ensuring that my variables are explicitly passed to the function in bplapply() because I am working on windows. I was having difficulties with the SnowParam() default not correctly evaluating function arguments whose values are derived from global environment. I followed the recommendations at #125 

To test out the code, I first ran:
```R 
prop = 0.2
num = 2
test_list = list(1,2,3,4,5,6,7,8,9,10)
test_df = data.frame(test_list) #2

dummyFUN2(ob = test_df, n = num, p = prop)
```
Which worked as expected and produced this output:
```
[[1]]
[1] 7 3

[[2]]
[1] 3 5
```

I then tried to use the function on a SingleCellExperiment object `sce` which has 20,601 columns in its counts assay:
```R 
prop = 0.2
num = 2

dummyFUN2(ob = sce, n = num, p = prop)
```
But I kept getting the following message:

<details><summary>Error Output</summary>
<p>

```
Loading required package: SingleCellExperiment
Loading required package: SummarizedExperiment
Loading required package: MatrixGenerics
Loading required package: matrixStats

Attaching package: ‘MatrixGenerics’

The following objects are masked from ‘package:matrixStats’:

    colAlls, colAnyNAs, colAnys, colAvgsPerRowSet, colCollapse, colCounts, colCummaxs, colCummins, colCumprods, colCumsums, colDiffs,
    colIQRDiffs, colIQRs, colLogSumExps, colMadDiffs, colMads, colMaxs, colMeans2, colMedians, colMins, colOrderStats, colProds,
    colQuantiles, colRanges, colRanks, colSdDiffs, colSds, colSums2, colTabulates, colVarDiffs, colVars, colWeightedMads,
[...]

PR #242, opened 2023-02-25
Force evaluation of the function argument

Jiefei-Wang:
My git skill is a little bit rusty, I pushed to the master branch by accident. I have reverted the change but I think we can do a force reset to clean it up.

This is a pull request to solve the test 3 example in https://github.com/Bioconductor/BiocParallel/issues/241#issuecomment-1444297564. I force R to evaluate all function arguments before serializing the function object. This should solve the issue introduced by the delayed assignment.

### DIFF
--- a/R/bpvalidate.R
+++ b/R/bpvalidate.R
@@ -124,6 +124,10 @@ setMethod("show", "BPValidate", function(object) {
         i <- vapply(unknown, function(x) {
             !exists(x, envir = env, inherits = FALSE)
         }, logical(1))
+        ## Force evaluation of the known arguments to
+        ## make sure they will be exported correctly
+        known <- unknown[-i]
+        for(nm in known) force(env[[nm]])
         unknown <- unknown[i]
         env <- parent.env(env)
     }
--- a/inst/unitTests/test_bpexportglobals.R
+++ b/inst/unitTests/test_bpexportglobals.R
@@ -42,3 +42,14 @@ test_bpexportglobals_bplapply <- function()
     current <- bplapply(1:2, function(i) getOption("BAR"), BPPARAM=param)
     checkIdentical("baz", unique(unlist(current)))
 }
+
+test_bpexportglobals_lazyEvaluation <- function(){
+    foo <- function(k){
+        param <- SnowParam(2L, exportglobals=TRUE)
+        bplapply(1:2, function(x){
+            k
+        }, BPPARAM = param)
+    }
+    k <- 1
+    checkIdentical(foo(k), list(1, 1))
+}


R/bpvalidate.R
```diff
@@ -124,6 +124,10 @@ setMethod("show", "BPValidate", function(object) {
         i <- vapply(unknown, function(x) {
             !exists(x, envir = env, inherits = FALSE)
         }, logical(1))
+        ## Force evaluation of the known arguments to
+        ## make sure they will be exported correctly
+        known <- unknown[-i]
+        lapply(known, function(x) env[[x]] <- env[[x]])
```
mtmorgan: use `force()` instead?

mtmorgan (commented):
Provide unit test(s)

Jiefei-Wang:
Thanks for your comment. I have added a unit test for it. This unit test will fail in the current devel version but will pass in this branch.

R/bpvalidate.R
```diff
@@ -124,6 +124,10 @@ setMethod("show", "BPValidate", function(object) {
         i <- vapply(unknown, function(x) {
             !exists(x, envir = env, inherits = FALSE)
         }, logical(1))
+        ## Force evaluation of the known arguments to
+        ## make sure they will be exported correctly
+        known <- unknown[-i]
+        lapply(known, function(x) force(env[[x]]))
```
mtmorgan: Since the return value is not used, does it make better sense to use a `for()` loop?

Jiefei-Wang: I also thought about this. Here is my blind spot. Will a for-loop be more efficient than an apply function? 

I know in practice this may not make any visible difference, but I was always told that we should avoid using for-loop if possible. Therefore, I chose `lapply` as it is more concise than the for-loop.

mtmorgan: The advice against `for` loops is partly historical, when R would grow a result one element at a time, resulting in quadratic scaling even in a simple loop

```
result <- integer()
for (i in 1:100000) result[i] <- i
```
Historically this code would run very slowly (like in the next code chunk). This has been addressed by a simple change. Instead of extending `result` by 1 each iteration, over-extend it by a small proportion larger than 1 each time more space is needed. This actually changes the performance from quadratic to logarithmic. The code above runs in about 1/20th of a second.

A `for` loop still scales poorly with the 'copy-and-append' pattern that is not infrequently seen
```
system.time({
    result <- integer()
    for (i in 1:100000) result <- c(result, i)
})
##    user  system elapsed
##  39.816   2.937  42.905
```

Yikes! why worry about this when `lapply()` etc do all the memory management?

But in the `force()` loop, there are no memory management issues associated with growing a result during the iteration, so no costs like this to a poorly formulated loop. 

It's very hard to analyze memory use, but `lapply()` will retain objects from each iteration and I guess increments the reference count on the object. If the result of each iteration are not used, then this is wasted memory; the `for()` loop might also allow more efficient garbage collection.

More subjectively, the `lapply()` without assignment somehow looks wrong to me; at some point in the future I sense that I'll return to this line of code, see that it 'does nothing' (no assignment) and think that I can delete it...

Jiefei-Wang: Thanks for the explanation. That makes much sense. I have updated the code to use the for loop.

R/bpvalidate.R
```diff
@@ -124,6 +124,10 @@ setMethod("show", "BPValidate", function(object) {
         i <- vapply(unknown, function(x) {
             !exists(x, envir = env, inherits = FALSE)
         }, logical(1))
+        ## Force evaluation of the known arguments to
+        ## make sure they will be exported correctly
+        known <- unknown[-i]
+        for(i in known) force(env[[i]])
```
mtmorgan: ...but now the for loop changes i? if this fails did your unit test capture this?

Jiefei-Wang: You are right, I have changed the variable `i` to `obj`

For some reason, I cannot run `devtools::check()` in the Ubuntu system.  I always get an error similar to 
```
ERROR in test_rng_lapply_bpredo: Error :   _R_CHECK_LIMIT_CORES_' environment variable detected, BiocParallel
  workers must be <= 2 was (3)
```
This should be related to the recent change in the worker number. However, what puzzled me is that I can manually run the auto-generated check file  `BiocParallel.Rcheck/BiocParallel-Ex.R` with no problem. `R CMD check` is also fine. The only way to reproduce it is to run `devtools::check()` or click the check button in RStudio. Therefore, I have no clue about why I can have this issue. Any thoughts or guess?

mtmorgan: `devtools::check()` tries to emulate the CRAN build environment, where `_R_CHECK_LIMIT_CORES_` is set. `R CMD check` (and the Bioconductor build system) do not set this variable. It can be set anyway, e.g.,
```
Sys.setenv("_R_CHECK_LIMIT_CORES_" = TRUE)
BiocGenerics::testPackage("BiocParallel", pattern = "test_rng.R")
BiocGenerics::testPackage("BiocParallel")
```
or
```
_R_CHECK_LIMIT_CORES_=TRUE R CMD check .
_R_CHECK_LIMIT_CORES_=TRUE Rscript BiocParallel.Rcheck/BiocParallel-Ex.R
```

mtmorgan:
merged in b10ee16b98717ab29d1bad73e629f4b08c68d4e6

PR #244, opened 2023-03-14
start TransientMulticoreParam, ensuring clean-up of environments

### LINKED ISSUE #243: MulticoreParam bplapply unable to restart upon interrupt
Hello,
coming from this [issue](https://github.com/carmonalab/UCell/issues/22), I have experienced in many instances - beyond that package - an odd behaviour.

Upon killing a function call that makes use of `bplapply` with `MulticoreParam()` I am no longer able to run the function again, as it throws the following error:

```
Error in d$value$value : $ operator is invalid for atomic vectors
```

I think a reproducible example could be running the code below and killing it with Ctrl-C, then trying to run it again:

```
a = BiocParallel::bplapply(seq_len(100), function(x) rnorm(x * 10^6, 0, 1), 
                                            BPPARAM = BiocParallel::MulticoreParam(workers = 2, progressbar = TRUE))
```

I need to restart the R session in order to use it. This happens both in RStudio and on the terminal, and it does not happen using `SnowParam()`. 

Should I specify additional parameters to the `bplapply` call with `MulticoreParam()` to avoid this behaviour?

<details>
  <summary>System and BiocParallel version</summary>

 R version 4.2.1 (2022-06-23)

 Platform: aarch64-apple-darwin20 (64-bit)

 Running under: macOS Monterey 12.5.1

 BiocParallel_1.32.5

</details>

mtmorgan:
- Closes #243

### DIFF
--- a/R/TransientMulticoreParam-class.R
+++ b/R/TransientMulticoreParam-class.R
@@ -7,7 +7,8 @@
 TransientMulticoreParam <-
     function(param)
 {
-    as(param, "TransientMulticoreParam")
+    param <- as(param, "TransientMulticoreParam")
+    bpstart(param)
 }
 
 .TRANSIENTMULTICOREPARAM_JOBNODE <- new.env(parent=emptyenv())
--- a/R/bpinit.R
+++ b/R/bpinit.R
@@ -29,10 +29,7 @@
             ## use TransientMulticoreParam when MulticoreParam has not
             ## started
             oldParam <- BPPARAM
-            BPPARAM <- as(BPPARAM, "TransientMulticoreParam")
-            ## BPPARAM will never get started as bpisup(BPPARAM) is always TRUE
-            if (!bpisup(oldParam))
-                .bpstart_set_rng_stream(BPPARAM)
+            BPPARAM <- TransientMulticoreParam(BPPARAM)
             on.exit({
                 .RNGstream(oldParam) <- .RNGstream(BPPARAM)
             }, TRUE, FALSE)


mtmorgan:
Hi @Jiefei-Wang not sure what you think of this?

TransientMulticoreParam uses `bpstart,TransientMulticoreParam` to 'clean' the state used, but the existing code did not call bpstart...

Also, the conditional test of `!bpisup(oldParam)` is irrelevant, since we only get to this point under that condition. And now that `bpstart()` is called, `.bpstart_set_rng_stream()` is also called.

`TransientMulticoreParam()` is not exported, and is not used elsewhere in the code.

Jiefei-Wang:
I haven't checked the code but I'm under the impression that we avoid using `bpstart` for `TransientMulticoreParam` on purpose. 

We want to reduce the amount of data transfer. Without calling `bpstatt`, the master will prepare everything the worker needs and directly copy itself. Therefore, the worker will inherit the data from the master. However, if we call `bpstart` in `bpinit`, the data is not ready yet(has not been divided into tasks) so we still need to use the SNOW connection to send the data to the worker. 

As you said, if TransientMulticoreParam uses bpstart,TransientMulticoreParam to 'clean' the state used, we might want to rename it and call it after parallelization.

mtmorgan:
Thanks Jiefei.

Actually TransientMulticoreParam avoids serialization during `.send_to()` where it forks a process directly and so doesn't serialize anything. I think it's OK to invoke `bpstart()` as part of the constructor.

Historically this comment https://github.com/Bioconductor/BiocParallel/pull/156#issue-1014566729 provides a justification and performance test case. Under Bioc-3.14 I see 

```
> f <-
+     function(param, y)
+ {
+     bplapply(1:100, function(x, y) sum(y), y = y, BPPARAM = param)
+ }
>
> y <- rnorm(1000000)
> param = MulticoreParam(tasks = 100)
> system.time(f(param, y)) # less than 1s
   user  system elapsed
  0.607   1.261   0.410
> param <- bpstart(MulticoreParam(tasks = 100))
> system.time(f(param, y)) # more than 4s
   user  system elapsed
  0.759   0.686   2.383
```
showing the benefit of TransientMulticoreParam. But in the current (patched or unpatched) devel I see
```
> param = MulticoreParam(tasks = 100)
> system.time(f(param, y)) # less than 1s
   user  system elapsed
  3.257   1.450   0.534
> param <- bpstart(MulticoreParam(tasks = 100))
> system.time(f(param, y)) # more than 4s
   user  system elapsed
  0.273   0.056   0.640
```
so I think the performance provided by TransientMulticoreParam has been addressed by other changes -- maybe in a different pull request it should just be removed to simplify the code base at very modest performance cost, or actually made even more performant (but that would probably be by improving the path to `.send_to()`, and benefit all Param?)?

PR #268, opened 2025-02-14
Use do.call() instead of lapply() in `bpoptions` (addressing issue #266)

### LINKED ISSUE #266: Can't find `bpprogressbar` unless `BiocParallel` is in `Depends`
Hello - hope it's ok to open an issue here, it didn't feel quite general enough for bioc-devel. 

We recently came across some (to us) unexpected behaviour of `BiocParallel`, and were hoping to understand whether there's just a gap in our understanding, or if there's an issue somewhere. 

More precisely, we wanted to add a progressbar inside `bplapply`, but even after importing `bpprogressbar` in our package, we get the error message `could not find function "bpprogressbar"` (e.g. when calling our functions in the vignette, when running `covr::report()`, or when running the tests as part of the `Check` in RStudio - however, just running the tests via the `Test` button in RStudio works). We could solve it by `Depend`ing on `BiocParallel`, so we have a workaround, but were wondering if this is the recommended way or if there are other options (I could not find this mentioned in the documentation, and as far as we could see our code looks very similar to the one provided in the examples of `bpoptions()`). 

I've made a minimal example package here: https://github.com/csoneson/TestBiocParallel 

The failing GitHub Actions run with `BiocParallel` in `Imports` is here: https://github.com/csoneson/TestBiocParallel/actions/runs/13218931852

And the passing one with `BiocParallel` in `Depends` (everything else equal): https://github.com/csoneson/TestBiocParallel/actions/runs/13223751571

Thanks in advance (and apologies if we overlooked something obvious)!
cc @mbstadler

csoneson:
Closes #266 

Just in case it saves some time, I put Martin's code fixing #266 in a PR.

### DIFF
--- a/R/bpoptions.R
+++ b/R/bpoptions.R
@@ -58,11 +58,10 @@
     ## find the common parameters both BPPARAM and BPOPTIONS
     paramOptions <- intersect(names(registeredOptions), optionNames)
     getterNames <- unlist(registeredOptions[paramOptions])
-    setNames(lapply(
-        getterNames,
-        do.call,
-        args = list(BPPARAM)
-    ), paramOptions)
+    result <- vector("list", length(getterNames))
+    for (i in seq_along(getterNames))
+        result[[i]] = do.call(getterNames[[i]], list(BPPARAM))
+    setNames(result, paramOptions)
 }
 
 ## value: BPOPTIONS


mtmorgan:
Thanks Charlotte!

